PR #39, opened 2016-05-10
Add autocomplete page. Closes #34

### LINKED ISSUE #34: Add autocomplete example
Currently, autocomplete is presented in InvoiceDemo Part 3. It should be added to KitchenSink as well.

joozek78:
//@warpech edit: For issue: https://github.com/StarcounterSamples/KitchenSink/issues/34


joozek78:
Please let me know what's the best approach to de-duplicating code in backend

warpech:
You can provide handlers in parent using the `Action` type. For example: https://github.com/StarcounterSamples/People/search?utf8=%E2%9C%93&q=ConfirmAction

src/KitchenSink/wwwroot/KitchenSink/style.css
```diff
@@ -94,4 +94,24 @@
 .kitchensink-toggle-button[disabled] {
     opacity: 0.5;
     cursor: not-allowed;
-}
\ No newline at end of file
+}
+
+ul.kitchensink-autocomplete {
```
warpech: This selector is not according to BEM. Should be: `.kitchensink-autocomplete` (without `ul`). See: http://starcounter.io/guides/web/avoiding-css-conflicts/

src/KitchenSink/wwwroot/KitchenSink/style.css
```diff
@@ -94,4 +94,24 @@
...
+    display: block;
+    background-color: white;
+    box-shadow: 0 2px 6px 0 rgba(0, 0, 0, 0.37);
+    list-style-type: none;
+    padding: 0px;
+    margin: 2px 0px;
+    position: absolute;
+}
+
+ul.kitchensink-autocomplete li {
```
warpech: This selector is not according to BEM. Should be: `.kitchensink-autocomplete__item` (without element selectors)

src/KitchenSink/wwwroot/KitchenSink/style.css
```diff
@@ -94,4 +94,24 @@
...
+    padding: 0px;
+}
+
+ul.kitchensink-autocomplete li a {
```
warpech: This selector is not according to BEM. Should be: `.kitchensink-autocomplete__link` (without element selectors)

warpech:
Reviewed, assigning back for some fixes of CSS selectors and potential usage of `Action`

joozek78:
It seems that hooking blur event works only in chrome. I'll some other way tomorrow

joozek78:
@warpech please take a look

warpech:
1. The Autocomplete Breadcrumb example does not work if you don't visit the BreadcrumbsPage first.
2. In `style.css` there is a redundant `}` character at line 122.
3. In the Places autocomplete, please add something for every letter of the alphabet. Otherwise it is hard to guess what values to type.
4. For both autocompletes, please add some summary: "Your current selection is: _______"

joozek78:
Why would I need a summary

joozek78:
@warpech

warpech:
I propose to change this line:

```cs
if (Db.SQL<long>("SELECT count (p) from Simplified.Ring3.Product p WHERE name = ?", name).First() == 0) {
```

to this:

```cs
if (Db.SQL<Product>("SELECT p FROM Simplified.Ring3.Product p WHERE Name = ? FETCH ?", name, 1).First == null) {
```

because that's a pattern we use in other places.

Other than that, it's good to merge!

warpech:
Looks good to merge!

PR #42, opened 2016-05-11
Tests for RedirectPage. #40

joozek78:


test/KitchenSink.Tests/Tests/RedirectPageTest.cs
```diff
@@ -0,0 +1,60 @@
...
+            FindButton("Redirect to Starcounter.io").Click();
+
+            // see https://github.com/PuppetJs/puppet-redirect/issues/3
+            if (browser == "firefox") {
+                _wait.Until(ExpectedConditions.AlertIsPresent());
+                driver.SwitchTo().Alert().Dismiss();
+            }
+
+            // redirecting to external page can take some time
+            _wait.Until(ExpectedConditions.UrlContains("http://starcounter.io/"));
```
aloscha: Maybe some timeout set?

joozek78: Timeout is set for whole test on line 22

aloscha:
I have problem with edge
image

test/KitchenSink.Tests/Tests/RedirectPageTest.cs
aloscha: Ahh, yeah :)

aloscha:
And second one in that test we don't test start from standalone page and then go to Redirect page by clicking "Redirect" in left menu.

joozek78:
It works on my machine. I'd try running on teamcity, but I don't have the access yet. Can you do it for me?

joozek78:
@ second comment - I don't see why we'd want want it. This behavior should be guaranteed by starcounter, and IMO shouldn't be tested here

warpech:
Maybe let's just have a one test to check if navigation works (for NavPage.html) and remove these menu checks in every other test spec?

aloscha:
@warpech create one test to navigate for all menu?

aloscha:
@joozek78 
Instead of Button.Click we use something like this:

`var button = FindButton("Redirect to Starcounter.io");`
`var action = new OpenQA.Selenium.Interactions.Actions(driver);`
`action.Click(button).Build().Perform();`

And this works for me

joozek78:
Isn't that equivalent? `.Click()` works on all browsers and is shorter without compromising clarity, why would I change it?

aloscha:
I test it on two version of MicrosoftEdgeDriver
- version 10.0.10586
- version 10.0.14316 (this is insider)
New version of insider is 10.0.14316
> Supported Operating System
> Windows 10 Tech Preview
> Windows Insiders Build 14342

I don't remember @warpech do you get installer (.msi) or standalone version (.exe). Because if second I think we need to move to installer.

But in neither first neither second version of MicrosoftEdgeDriver .click doesn't work for me.

How it looks like in TeamCity?

warpech:
I have version 10.0.14279.1000, which seems not available on the download page anymore. 

Could you check what version works the best? Then please put that version number and a link to it on the page http://starcounter.io/guides/web/acceptance-testing-with-selenium/#running-in-multiple-browsers? I have sent you an invitation to WordPress.

warpech:
For me the tests fail in Edge as well, with this error message: 

```txt
Test Name:	ClickingOnFruitShouldChangeUrlAndText
Test FullName:	KitchenSink.Tests.Tests.RedirectPageTest("edge").ClickingOnFruitShouldChangeUrlAndText
Test Source:	c:\www\StarcounterSamples\KitchenSink\test\KitchenSink.Tests\Tests\RedirectPageTest.cs : line 28
Test Outcome:	Failed
Test Duration:	0:00:04.132

Result StackTrace:	
at OpenQA.Selenium.Remote.RemoteWebDriver.UnpackAndThrowOnError(Response errorResponse)
at OpenQA.Selenium.Remote.RemoteWebDriver.Execute(String driverCommandToExecute, Dictionary`2 parameters)
at OpenQA.Selenium.Remote.RemoteWebElement.Click()
at KitchenSink.Tests.Tests.RedirectPageTest.ClickingOnFruitShouldChangeUrlAndText() in c:\www\StarcounterSamples\KitchenSink\test\KitchenSink.Tests\Tests\RedirectPageTest.cs:line 30
Result Message:	System.InvalidOperationException : Element is obscured (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 2 milliseconds
Build info: version: '2.52.0', revision: '4c2593c', time: '2016-02-11 19:06:42'
System info: host: 'DESKTOP-ITUP0CI', ip: '192.168.1.79', os.name: 'Windows 10', os.arch: 'x86', os.version: '10.0', java.version: '1.8.0_65'
Driver info: org.openqa.selenium.edge.EdgeDriver
Capabilities [{acceptSslCerts=true, browserVersion=25.10586.0.0, platformVersion=10, browserName=MicrosoftEdge, takesScreenshot=true, pageLoadStrategy=normal, takesElementScreenshot=true, platformName=windows, platform=ANY}]
Session ID: 15E4B378-3460-464F-9E96-2DDEAC1A66D1
```

Switching to this implementation of `Click` method provided by @aloscha fixes that problem in Edge for me.

Guys, in KitchenSink Test there is a mess in the ways how the `Click` event is triggered. @joozek78 can you find one method that works best, put it to a helper class and use the same method in all of the tests?

joozek78:
@warpech by 'all of the tests' do you mean all across all projects? That would require separate project to hold common code, so I assumed you only mean all in KitchenSink. Please take a look

aloscha:
For me ok, all tests pass.
Nice one with extension :+1:

PR #45, opened 2016-05-12
Simplify Selenium instructions even more

warpech:
Just add a link to http://starcounter.io/guides/web/acceptance-testing-with-selenium/#running-in-multiple-browsers

@joozek78 what do you think of this change? I think it is the simplest, and can be pasted in other READMEs as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,24 +56,6 @@ To:
 
 ## Testing
 
-### Adding tests for your project?
-
-Go to Tools → NuGet Packet Manager → Packet Manager Console.
-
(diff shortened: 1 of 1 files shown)


joozek78:
Yeah, I didn't know that we had one central place with instruction. It's definitely better to reference it. But shouldn't we offload all of the 'testing' chapter there?

warpech:
> But shouldn't we offload all of the 'testing' chapter there?

I will leave it as it is right now, because the testing instructions might be project specific.

Thanks, merging

PR #56, opened 2016-06-30
Pattern for input validation #6

aloscha:


warpech:
Looking good! But can you please add tests?

aloscha:
@warpech just added

PR #64, opened 2016-09-29
TeamCity change in 'Apps :: KitchenSink :: Versioned Settings Test' p…

joozek78:
…roject: Synchronization with own VCS root is enabled


warpech:
@joozek78 should this be closed/removed?

PR #69, opened 2016-10-21
Update for double values in JsonByExample for pnext versions of Starcounter.

chrhol:
Due to the changes in pnext when we switched over to using NewtonSoft.Json as parser for JsonByExample, double values cannot be distinguished directly in Json and needs to be specified in code-behind (https://github.com/Starcounter/Starcounter/issues/3811#issuecomment-243408842).

Documented here: http://starcounter.io/guides/json/#supported-datatypes

This change will not disrupt anything in develop, setting this will just not make any difference there.

(A side-note: the pnext version of KitchenSink is currently failing due to newtonsoft is missing from the installer/package. A new nightly version is currently being built and will finish in a couple of hours)

CC @joozek78


chrhol:
Nightly pnext is finished now, and running KitchenSink with this branch succeeds.

PR #71, opened 2016-10-24
Use implicit instead of explicit wait in tests

warpech:
Implements: https://github.com/StarcounterSamples/KitchenSink/issues/70

**Let's discuss pros and cons before merging**

@miyconst, @joozek78, @aloscha, @Mnosek, @ubbeK, @Joltrix WDYT?

### DIFF
--- a/test/KitchenSink.Tests/Tests/AutocompletePageTest.cs
+++ b/test/KitchenSink.Tests/Tests/AutocompletePageTest.cs
@@ -88,6 +88,7 @@ namespace KitchenSink.Tests.Tests
             var placesSearchbox = driver.FindElement(PlacesSearchSelector);
             placesSearchbox.SendKeys("po");
             placesSearchbox.SendKeys(Keys.Tab);
+            this.WaitUntil(d => d.FindElements(FoundPlacesSelector).Count != 0);
             driver.FindElement(ProductsSearchSelector).Click();
             this.WaitUntil(d => d.FindElements(FoundPlacesSelector).Count == 0);
             Assert.IsEmpty(driver.FindElements(FoundPlacesSelector));
--- a/test/KitchenSink.Tests/Tests/AutocompletePageTest.cs
+++ b/test/KitchenSink.Tests/Tests/AutocompletePageTest.cs
@@ -52,6 +52,7 @@ namespace KitchenSink.Tests.Tests
             var countryToPick = "Poland";
             AssertElements(FoundPlacesSelector, countryToPick, "Portugal");
             driver.FindElements(FoundPlacesSelector).First(el => el.Text == countryToPick).Click();
+            this.WaitUntil(d => driver.FindElements(FoundPlacesSelector).Count == 0);
             Assert.AreEqual(countryToPick, driver.FindElement(PlacesSearchSelector).GetAttribute("value"),
                 "Search textbox has invalid content");
             Assert.AreEqual("Capital of Poland is Warsaw",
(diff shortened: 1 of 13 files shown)


warpech:
This requires current Selenium (`Shared\Distrib\selenium3-windows10` in Dropbox)

ubbeK:
I havent tried this out yet but what I saw from your `https://github.com/StarcounterSamples/KitchenSink/commit/1b8c962a0a0f357b5ab02f400174cf63822d8327` is a lot less code which usually means easier test development/scripting and especially easier maintenance which has a huge importance when running regression tests. A test that always fails and nobody knows why is just a waste resources.

Resources for tests that takes a bit longer should not be a problem. If it is, then we should use some open source CI server. How much longer are we talking about?

miyconst:
@warpech if the test are really passing with the implicit waiting, then I am all for it. Unfortunately I don't have time to check it myself at the moment.

I see that you have reverted your changes, why is so?

>Drawback: this extends the test execution time

We have to finally add parallel execution, if not in one browser, then in different browsers for sure.

warpech:
> I see that you have reverted your changes, why is so?

I have reverted some of the changes. It turns out that Selenium's implicit waiting only waits for elements that do not exist yet. If an element exists, it proceeds with the old value.

> @warpech if the test are really passing with the implicit waiting, then I am all for it. Unfortunately I don't have time to check it myself at the moment.

The tests are passing locally but not on the build agent. I am investigating.

miyconst:
>The tests are passing locally but not on the build agent. I am investigating.

Yeah, same thing happened for me with the parallel test execution.

warpech:
Build is now fixed. The problem turned out to be with Firefox, not with our tests. For me this can be merged.

miyconst:
Cool. Let's merge it.

test/KitchenSink.Tests/Tests/AutocompletePageTest.cs
```diff
@@ -29,19 +29,15 @@ public AutocompletePageTest(string browser) : base(browser)
...
             driver.Navigate().GoToUrl(baseURL + "/Autocomplete");
-            this.WaitUntil(ExpectedConditions.PresenceOfAllElementsLocatedBy(By.CssSelector("html body puppet-client")));
-            this.WaitUntil(ExpectedConditions.PresenceOfAllElementsLocatedBy(PlacesSearchSelector));
         }
 
         [Test]
         public void FillStarExpectAllItemsShowUp()
         {
             driver.FindElement(PlacesSearchSelector).SendKeys("*");
-            WaitForElementsToLoad(FoundPlacesSelector);
```
joozek78: shouldn't we still wait for this explicitly? Starcounter has to do its work here

warpech: I believe we don't have to wait for this explicitly. 

Method `WaitForElementsToLoad` was removed in the PR, because it is no longer needed to wait for unexisting element to appear. 

When Selenium executes the following line: 

```cs
Assert.AreEqual(7, driver.FindElements(FoundPlacesSelector).Count);
```

It will automatically wait up to 5 seconds until `Count` becomes higher than 0.

test/KitchenSink.Tests/Tests/AutocompletePageTest.cs
```diff
@@ -94,7 +87,6 @@ public void FillPlaceNameThenBlurExpectHintsGone()
             }
             var placesSearchbox = driver.FindElement(PlacesSearchSelector);
             placesSearchbox.SendKeys("po");
-            WaitForElementsToLoad(FoundPlacesSelector);
```
joozek78: to be restored.
If you blur immediately (without waiting for FoundPlaces to appear) you won't test that FoundPlaces disappear (they could have never appeared)

warpech: Totally, agree. I have pushed a fix.

warpech:
Thanks for the review, merging

PR #89, opened 2017-02-08
Separate Dropdown and Autocomplete data

un-tone:
For https://github.com/StarcounterSamples/KitchenSink/issues/86


src/KitchenSink/BreadcrumbPage.json.cs
```diff
@@ -22,13 +22,8 @@ protected override void OnData()
...
 
-            if (BreadcrumbTestData.Exists())
-            {
-                Db.Transact(() => { BreadcrumbTestData.DeleteAll(); });
-            }
-            BreadcrumbTestData.Create();
-
-            var treeItem = Db.SQL<TreeItem>("SELECT i FROM TreeItem i WHERE Name = ?", "Milk").First;
+            var treeItem = Db.SQL<TreeItem>("SELECT i FROM TreeItem i WHERE Name = ?", "Milk").First
+                ?? Db.SQL<TreeItem>("SELECT i FROM TreeItem i FETCH ?", 1).First;
```
warpech: Why is line 26 needed?

warpech: Why is line 26 needed?

un-tone: If a user changes the item with name "Milk" to any other name (e.g. "Milk2") the query `Db.SQL<TreeItem>("SELECT i FROM TreeItem i WHERE Name = ?", "Milk").First` will return `null` and in the following code the NullReferenceException will be thrown.
Doing this query `Db.SQL<TreeItem>("SELECT i FROM TreeItem i FETCH ?", 1).First` eliminates it

warpech: Good catch! But still if I change the name to "Milk2" and reload the page, I am getting null reference exception. Can you reproduce that?

un-tone: Sure, thanks. I missed a filtering by Parent field. Fixed.

src/KitchenSink/BreadcrumbPage.json.cs
```diff
@@ -22,13 +22,8 @@ protected override void OnData()
...
-            if (BreadcrumbTestData.Exists())
-            {
-                Db.Transact(() => { BreadcrumbTestData.DeleteAll(); });
-            }
-            BreadcrumbTestData.Create();
-
-            var treeItem = Db.SQL<TreeItem>("SELECT i FROM TreeItem i WHERE Name = ?", "Milk").First;
+            var treeItem = Db.SQL<TreeItem>("SELECT i FROM TreeItem i WHERE Name = ?", "Milk").First
+                ?? Db.SQL<TreeItem>("SELECT i FROM TreeItem i WHERE Parent IS NOT NULL FETCH ?", 1).First;
             SetActiveItem(treeItem);
```
warpech: Now it works good, but maybe let's forget about "Milk" and simplify it by just calling:

```
var treeItem = Db.SQL<TreeItem>("SELECT i FROM TreeItem i WHERE Parent IS NOT NULL FETCH ?", 1).First;
```

un-tone: OK, removed, please check.

KmaTTi89:
There is a problem while merging with KitchenSink.csproj

un-tone:
@KmaTTi89, what is the problem? Could you say how can I check it?

KmaTTi89:
@un-tone The problem is, I guess, your changes are based on dev branch and my current branch where i am testing has a lot of changes.
Could you try to merge your branch with my: new_tests_project

un-tone:
@KmaTTi89, I fixed the formatting of the KitchenSink.csproj, please check

warpech:
@KmaTTi89 I think that we need to decide to one target branch. Would you like this to be merged to `develop` or `new_tests_project`?

warpech:
I think that the correct flow is to merge to `develop` and then @KmaTTi89 solves conflicts in his branch since he owns that branch.

warpech:
Ok, then it is fine by me!

@un-tone maybe you can propose strategy to close this PR. Since @KmaTTi89 has merged from one feature branch to another feature branch, it should be enough to change the PR target from `develop` to `new_tests_project` to merge without conflicts.

un-tone:
@warpech, in a standard case I would target these PRs to `develop`. But since these issues are related directly to the test covering issues I assume it more correct to target the PRs to the `new_tests_project` branch.

un-tone:
Please confirm re-targeting of the PR.

warpech:
I confirm.

un-tone:
Closing. Changes from the `feature/Separate-Dropdown-Autocomplete-data` branch is already merged with `new_tests_project`

PR #99, opened 2017-02-28
Parallelizable test

KmaTTi89:
- update tests to run parallel
- update nunit
- update webdriver
- disable screenshot maker due firefox bug

### DIFF
--- a/test/KitchenSink.Tests/KitchenSink.Tests.csproj
+++ b/test/KitchenSink.Tests/KitchenSink.Tests.csproj
@@ -37,18 +37,18 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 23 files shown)


warpech:
@miyconst Could you please review? Does it work like you wanted https://github.com/StarcounterSamples/KitchenSink/pull/62 to work?

miyconst:
@warpech you never stop working, even at the airport :). Will check it as soon as I get home.

miyconst:
Yes. This is something I wanted. @KmaTTi89 could you please share numbers? How much time do we save by executing it in parallel? Unfortunately I don't have Selenium environment at my home PC to check.

warpech:
@KmaTTi89 wrote me on Slack that it decreased testing time from 5 minutes to 50 seconds.

warpech:
On my machine the running time is decreased from 6m15.711s to 3m52.919s.

Measured using `time cmd //c "test.bat"` in Git Bash.

KmaTTi89:
Old build on TeamCity: about 5 minutes
New build on TeamCity with Parallel tests: about 1 and a half minute

PR #101, opened 2017-03-01
Uniform batch script

KmaTTi89:
- edit run.bat to check if script is running on TC
- edit test.bat to not close after running the tests

### DIFF
--- a/run.bat
+++ b/run.bat
@@ -1,8 +1,3 @@
 @echo off
 
-:: If we are not running on build server, stop the db to stop the running apps
-IF NOT "%TEAMCITY_VERSION%"=="True" (
-    staradmin stop db default
-)
-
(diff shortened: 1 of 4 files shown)


run.bat
```diff
@@ -1,3 +1,8 @@
 @echo off
 
-START /W star.exe --resourcedir="src\KitchenSink\wwwroot" "bin\Debug\KitchenSink.exe"
\ No newline at end of file
+:: If we are not running on build server, stop the db to stop the running apps
+IF NOT "%TEAMCITY_VERSION%"=="True" (
+    staradmin stop db default
+)
```
warpech: Normally, unless I am testing, I actually **want** multiple apps to be running side-by-side. So this should not be here.

KmaTTi89: Fixed

PR #103, opened 2017-03-10
Fixes and updates

KmaTTi89:
- fix for test.bat, don't use "pause" command when running script from console
- updated nunit console runner
- updated webdriver lib

### DIFF
--- a/test.bat
+++ b/test.bat
@@ -4,13 +4,11 @@ set interactive=1
 echo %cmdcmdline% | find /i "%~0" >nul
 if not errorlevel 1 set interactive=0
 
-rem now I can use %interactive% anywhere
+rem If we are in interactive mode (batch file not started from the command line), pause at the exit
 
 IF NOT EXIST "%~dp0packages\NUnit.ConsoleRunner.3.6.1\" (ECHO Error: Cannot find NUnit Console Runner. Build the project to restore the packages && PAUSE && EXIT /B 1)
-%~dp0packages\NUnit.ConsoleRunner.3.6.1\tools\nunit3-console.exe %~dp0test\KitchenSink.Tests\bin\Debug\KitchenSink.Tests.dll --noheader --params Browsers=Chrome,Firefox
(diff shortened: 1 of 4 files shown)


test.bat
```diff
@@ -1,6 +1,16 @@
 @echo off
 
-IF NOT EXIST "%~dp0packages\NUnit.ConsoleRunner.3.6.0\" (ECHO Error: Cannot find NUnit Console Runner. Build the project to restore the packages && PAUSE && EXIT /B 1)
-%~dp0packages\NUnit.ConsoleRunner.3.6.0\tools\nunit3-console.exe %~dp0test\KitchenSink.Tests\bin\Debug\KitchenSink.Tests.dll --noheader --params Browsers=Chrome,Firefox
-pause
+set interactive=1
+echo %cmdcmdline% | find /i "%~0" >nul
+if not errorlevel 1 set interactive=0
+
+rem now I can use %interactive% anywhere
```
warpech: This comment is redundant here. Can you change it to something more meaningful like `If we are in interactive mode (batch file not started from the command line), pause at the exit`

test.bat
```diff
@@ -1,6 +1,16 @@
...
+
+IF NOT EXIST "%~dp0packages\NUnit.ConsoleRunner.3.6.1\" (ECHO Error: Cannot find NUnit Console Runner. Build the project to restore the packages && PAUSE && EXIT /B 1)
+%~dp0packages\NUnit.ConsoleRunner.3.6.1\tools\nunit3-console.exe %~dp0test\KitchenSink.Tests\bin\Debug\KitchenSink.Tests.dll --noheader --params Browsers=Chrome,Firefox
+
+if _%1_==__ (
```
warpech: The above line looks redundant. See: http://stackoverflow.com/questions/2309968/what-are-1-and-2-in-batch-files

test.bat
KmaTTi89: Changed

test.bat
KmaTTi89: Fixed

PR #104, opened 2017-03-14
Tests fix for new sc version

KmaTTi89:
In Starcounter version 2.3 kitchensink display some elements in #shadow-root.
This is a fix for tests on google chrome for data grid page and upload file page.

### DIFF
--- a/test/KitchenSink.Tests/Ui/BasePage.cs
+++ b/test/KitchenSink.Tests/Ui/BasePage.cs
@@ -93,7 +93,12 @@ namespace KitchenSink.Tests.Ui
         public IWebElement ExpandShadowRoot(IWebElement shadowRootElement)
         {
             IWebElement shadowTreeParent = (IWebElement)((IJavaScriptExecutor)Driver)
-            .ExecuteScript("return arguments[0].shadowRoot", shadowRootElement);
+           .ExecuteScript("return arguments[0].shadowRoot", shadowRootElement);
+            if (shadowTreeParent == null)
+            {
+                //Shadow DOM not supported, fall back to DOM
+                return shadowRootElement;
+            }
             return shadowTreeParent;
         }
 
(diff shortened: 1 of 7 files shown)


test/KitchenSink.Tests/Ui/BasePage.cs
```diff
@@ -90,6 +90,13 @@ public void ClickOn(IWebElement elementName, int seconds = 10)
             element.Click();
         }
 
+        public IWebElement ExpandShadowRoot(IWebElement shadowRootElement)
+        {
+            IWebElement shadowTreeParent = (IWebElement)((IJavaScriptExecutor)Driver)
+            .ExecuteScript("return arguments[0].shadowRoot", shadowRootElement);
+            return shadowTreeParent;
+        }
```
warpech: I understand that here you are checking if the definition of Polymer element has been loaded and rendered. You developed a clever way to do it. 

With the below change, the method will be equally usable in all browsers (Chrome and not Chrome). With it you can remove the special checks for not Chrome in your tests:

```cs
        public IWebElement ExpandShadowRoot(IWebElement shadowRootElement)
        {
            IWebElement shadowTreeParent = (IWebElement)((IJavaScriptExecutor)Driver)
            .ExecuteScript("return arguments[0].shadowRoot", shadowRootElement);
            if (shadowTreeParent == null)
            {
                //Shadow DOM not supported, fall back to DOM
                return shadowRootElement;
            }
            return shadowTreeParent;
        }
```

KmaTTi89: Ok thanks, I will fix code with your suggestion.
Currently only chrome support shadow dom and tests fail only on it

warpech: > Currently only chrome support shadow dom and tests fail only on it

Since we know that all browsers will implement it eventually, we should prepare for it. Otherwise the tests may start failing after a browser upgrade.

For the above reason, it is considered a good practice to do **feature detection** instead of **browser detection**.

PR #108, opened 2017-03-16
Use "Trigger" suffix on button handlers

Mackiovello:


alshakero:
@Mackiovello your work is always super elegant. Branch names, commits messages, everything is so damn on point.

Mackiovello:
Thank you @alshakero! I appreciate it

PR #119, opened 2017-04-24
Teamcity script refactoring

KmaTTi89:
Make Teamcity.ps1 more generic

### DIFF
--- a/tools/Teamcity.ps1
+++ b/tools/Teamcity.ps1
@@ -1,29 +1,38 @@
-Param ($checkoutdir, $nunitversion, $browsersToRun, $testedApp, [String[]] $appsToRun, [String[]] $helpersToRun, $testsPath)
+Param
+(
+	[Parameter(Mandatory=$true)][string] $checkoutDir, 
+	[Parameter(Mandatory=$true)][string] $nunitVersion, 
+	[Parameter(Mandatory=$true)][string] $browsersToRun, 
+	[Parameter(Mandatory=$true)][string] $testedApp, 
+	[Parameter(Mandatory=$true)][string[]] $appsToRun, 
+	[Parameter(Mandatory=$false)][string[]] $helpersToRun, 
+	[Parameter(Mandatory=$false)] $testsPath
+)
 
-$StarCounterDir = "$checkoutdir\sc"
-$StarCounterWorkDirPath = "$StarCounterDir\starcounter-workdir"
-$StarCounterRepoPath = "$StarCounterWorkDirPath\personal"
-$StarCounterConfigPath = "$StarCounterDir\Configuration"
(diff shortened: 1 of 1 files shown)


warpech:
@joozek78 could you pls review this? The purpose of this change it to make it possible to use the same script in all the projects. The apps to build and tests to build & run should be passed to it as input params.

tools/Teamcity.ps1
```diff
@@ -1,56 +1,91 @@
...
+{
+	$fileContent = "<?xml version=`"1.0`" encoding=`"UTF-8`"?><service><server-dir>$StarcounterRepoPath</server-dir></service>"
+	New-Item -Path $StarcounterConfigPath -Name personal.xml -ItemType "file" -force -Value $fileContent | Out-Null
+}
 
-$NunitConsoleRunnerExePath = "$checkoutdir\KitchenSink\packages\NUnit.ConsoleRunner.$nunitversion\tools\nunit3-console.exe"
-$NunitArg = "$KitchenSinkTestsPath --noheader --teamcity --params Browsers=$browsersToRun"
+Function createRepo()
+{
+	Start-Process -FilePath $StarExePath -ArgumentList "`@`@createrepo $StarcounterWorkDirPath" -NoNewWindow -Wait
```
joozek78: why is only first `@` escaped?

tools/Teamcity.ps1
```diff
@@ -1,56 +1,91 @@
...
-$StarAdminExePath = "$StarCounterDir\staradmin.exe"
+Function runApps($apps, $source)
+{
+	foreach ($app in $apps)
+	{
+		$AppWWWPath = "$checkoutDir\$testedApp\$source\$app\wwwroot"
+		$AppExePath = "$checkoutDir\$testedApp\$source\$app\bin\Debug\$app.exe"
+		$AppArg = "--resourcedir=$AppWWWPath $AppExePath"
+		
+		$process = Start-Process -FilePath $StarExePath -ArgumentList $AppArg -PassThru -NoNewWindow
```
joozek78: why not just `-Wait`?

tools/Teamcity.ps1
KmaTTi89: I can see that second one is escaped too but maybe only on my monitor
The PowerShell escape character is the grave-accent(`)

tools/Teamcity.ps1
KmaTTi89: I can provide id of the process to wait

tools/Teamcity.ps1
joozek78: I get it now :) It looks like the first one is "quoted", but actually both are preceded with "`"

PR #120, opened 2017-04-25
Added test for Markdown page

KmaTTi89:


warpech:
@KmaTTi89 please create cross links between PRs and Issues.

Implements: https://github.com/StarcounterApps/KitchenSink/issues/26

@tomalec could you pls review the test?

KmaTTi89:
It is not my fault that someone use header text as id in html tag. I use id to find element I want to work on

warpech:
@KmaTTi89 I disagree. In this case you actually MUST check that "This is a strucured text" is in a `<h1>` tag, because this is what Markdown conversion does.

Checking id of an element doesn't tell us that Markdown was correctly rendered as HTML.

Btw pls fix a typo in: This is a **strucured** text

test/KitchenSink.Tests/Test/SectionString/MarkdownPageTest.cs
```diff
@@ -0,0 +1,35 @@
...
+            _mainPage = new MainPage(Driver).GoToMainPage();
+            _markdown = _mainPage.GoToMarkdownPage();
+        }
+
+        [Test]
+        public void MarkdownPage_CheckPreviewText()
+        {
+            WaitUntil(x => _markdown.CheckPreviewVisible());
+
+            Assert.AreEqual("This is a structured text", _markdown.GetPreviewText());
```
tomalec: I'd rename it to something like `_GetHeaderText` to indicate that you are checking only `# This is a structured text`, not the entire markdown preview  with 
```md
# This is a structured text
  		  
It supports **markdown** *syntax*."
```

test/KitchenSink.Tests/Ui/SectionString/MarkdownPage.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        public MarkdownPage(IWebDriver driver) : base(driver)
+        {
+            PageFactory.InitElements(Driver, this);
+        }
+
+        public bool CheckPreviewVisible()
+        {
+            var shadowRoot = ExpandShadowRoot(Driver.FindElement(By.XPath("//juicy-markdown")));
+            return shadowRoot.FindElement(By.TagName("h1")).Displayed; //BUG in CHROME b.getElementsByTagName is not a function
```
tomalec: I would rather wait for the element to be visible, not for it's content to be valid, as it's in fact actual assertion hidden in `WaitUntil`, sure it will fail if the element is not working correctly, but the failure message will not be helpful.

test/KitchenSink.Tests/Test/SectionString/MarkdownPageTest.cs
KmaTTi89: I can change the name of function

test/KitchenSink.Tests/Ui/SectionString/MarkdownPage.cs
KmaTTi89: What is the problem with this lines?

tomalec: Correct me if I'm wrong, but in case of `<juicy-markdown>` fails to render `<h1>`, your test will fail with "timeout" error, not with "h1 is not displayed" error.

While testing at https://github.com/StarcounterApps/KitchenSink/pull/120/files#diff-0bc3408b5756f5569d21e5dcd20c719bR30 I'd just wait for the `juicy-markdown` element.

Then assert at https://github.com/StarcounterApps/KitchenSink/pull/120/files#diff-0bc3408b5756f5569d21e5dcd20c719bR32
that 
- `<h1>` is displayed
- `<h1>` contains given text `This is a structured text`
.

KmaTTi89: Check now

tomalec: Thanks, LGTM

Issue #76, opened 2016-11-07
Chart does not render on polyfilled browsers


tomalec:
### Steps to reproduce

1. Run KitchenSink
2. Open [/KitchenSink/Chart](http://localhost:8080/KitchenSink/Chart)

### Expected result
Chart should be shown

### Actual result
...it's not

### Technical information

Error is thrown to console:
```log
15:07:00.297 TypeError: this.scale is undefined
.draw()Chart.js:2781
helpers.animationLoop/animationFrame()Chart.js:697
1Chart.js:2781:4
```

##### Involved apps' version

 - StarcounterSamples/KitchenSink@4f6309355854cf298a1decba14c5ea298dae35b5 (current HEAD of develop)
 
##### Browsers affected

- [ ] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 9
- [?] Safari 8
- [x] IE 11

### Screenshot
![image](https://cloud.githubusercontent.com/assets/17435/20060575/269e79ce-a4fc-11e6-8ef6-f34c3e4f28a4.png)

warpech:
@PatrykSzwer could you pls take a look at this?

PatrykSzwer:
It's still reproducing. Trying to fix this now without results yet.

warpech:
Maybe upgrading to the current version helps?

PatrykSzwer:
I've made PR (#127) with fix on that. I've rebuild entire way that chart on chartPage works, by creating custom element.

PatrykSzwer:
Fixed in #127. Closing.

PR #127, opened 2017-05-08
76 chart on polyfilled browsers

PatrykSzwer:
#76

### DIFF
--- a/src/KitchenSink/wwwroot/KitchenSink/ChartPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/ChartPage.html
@@ -3,29 +3,56 @@
 <template>
     <template is="dom-bind">
         <p slot="kitchensink/chartpage-description" class="lead">JSON's <code>array</code> can be used to create a chart with a helper custom element, e.g. <a href="http://robdodson.github.io/chart-elements">chart-elements</a>.</p>
+        
+        <kitchensink-chart slot="kitchensink/chartpage-chart" temperatures="{{model.Temperatures}}" labels="{{model.Labels}}"></kitchensink-chart>
+    </template>
+</template>
 
-        <chart-line slot="kitchensink/chartpage-chart" id="chrt"></chart-line>
+<dom-module id="kitchensink-chart">
+    <template>
+        <h1>hello</h1>
+        <chart-line id="chrt" data="[[chartData]]"></chart-line>
     </template>
     <script>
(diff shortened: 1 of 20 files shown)


src/KitchenSink/wwwroot/KitchenSink/ChartPage.html
```diff
@@ -3,29 +3,55 @@
...
     <script>
-        var template = document.currentScript.previousElementSibling;
-        template.addEventListener('dom-change', function () {
-            var chart = document.getElementById("chrt");
+        Polymer({
+            is: "kitchensink-chart",
+            properties: {
+                temperatures: {
+                    type: Array,
+                    observer: '_temperaturesChanged'
```
tomalec: Do you need to separate them, as in both cases you are resetting entire `this.chartData`?
Maybe you can call it `resetChartData` instead of `updateChartData` on call it for both observers?

Or even - if you don't plan to support changing them separately - you can bind `{{model}}` to single property, to reduce ping-pong in data-binding.

Other than that, I think original code didn't work because `document.currentScript` is polyfilled there as `document._currentScript` so maybe it would be fixed with single liner.

I find custom-element fix quite clean one, and really like it as it's nice show-case of how our `imported-template` works and the fact that it's a great place and way to define new custom elements.

However, I doubt weather it's worth to introduce performance overhead in defining, polyfilig and stamping custom element to use it just once. What may also be important, it may increase the cognitive load and be little bit confusing, that you need to define custom element to use custom element.

I would ping the ball to @warpech as I'm not sure what ChartPage was intended to show-case.

warpech: I think that it totally doesn't matter how we do it. The `ChartPage` is supposed to teach JSON data binding, not to teach Polymer. It shows two things that primitive data types (such as array) might be used to show a graph -- and how to do that in a declarative way. In future I think that this page should be expanded to push data points into the graph.

So @tomalec are you suggesting to just replace `observer: '_temperaturesChanged'` with  `observer: 'updateChartData'`?

tomalec: yes :)

PatrykSzwer: Done. Let me know if It's everything, then I'll merge and close.

PR #129, opened 2017-05-15
Pagination improvement

Mackiovello:
Minor improvement to the pagination page. See commit messages

### DIFF
--- a/src/KitchenSink/PaginationPage.json.cs
+++ b/src/KitchenSink/PaginationPage.json.cs
@@ -45,6 +45,7 @@ namespace KitchenSink
             {
                 CurrentOffset = CurrentOffset + EntriesPerPage;
             }
+
             GetNewPage();
         }
 
@@ -67,6 +68,7 @@ namespace KitchenSink
             {
                 CurrentOffset = TotalEntries - remainder;
             }
+
             GetNewPage();
         }
 
@@ -132,6 +134,7 @@ namespace KitchenSink
                     }
                 }
             }
+
             DisableLast = CurrentPage == TotalPages;
             DisableFirst = CurrentPage == 1;
         }
--- a/src/KitchenSink/PaginationPage.json.cs
+++ b/src/KitchenSink/PaginationPage.json.cs
@@ -1,4 +1,5 @@
 using Starcounter;
+using System.Linq;
 
 namespace KitchenSink
 {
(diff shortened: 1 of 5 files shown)


src/KitchenSink/PaginationPage.json.cs
```diff
@@ -16,143 +17,137 @@ protected override void OnData()
...
+            CreateNavButtons();
+            SetPageEntries();
         }
 
-        // Changes the number of entries per page depending on input and sets the page to 1
+        public long TotalEntries = Db.SQL<long>("SELECT COUNT(e) FROM KitchenSink.Book e").FirstOrDefault();
+
         void Handle(Input.EntriesPerPage action)
         {
-            this.CurrentOffset = 0;
```
PatrykSzwer: Is somewhere written that we should/shouldn't use `this.` on fields that are from class json file? I couldn't find it in guidelines or anywhere else.

src/KitchenSink/PaginationPage.json.cs
```diff
@@ -16,143 +17,137 @@ protected override void OnData()
...
+            long remainder = TotalEntries % EntriesPerPage;
+            if (remainder == 0)
+            {
+                CurrentOffset = TotalEntries - EntriesPerPage;
+            }
+            else
+            {
+                CurrentOffset = TotalEntries - remainder;
+            }
             GetNewPage();
```
PatrykSzwer: As we are in improvements I suggest that we add free space after closing braces when there is another lane after (method call, setting variable etc.). IMO It's more readable and clear. Tell me what do You think about it.

src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
```diff
@@ -2,13 +2,17 @@
...
     <template is="dom-bind">
+        <p slot="kitchensink/html-header" class="lead">Pagination can be implemented in the code-behind by using <code>FETCH</code> to determine how many entries there should be on every page and <code>OFFSET</code> to retrieve the right entries</p>
+        <div class="kitchensink-pagination-page-controls">
+            <juicy-select slot="kitchensink/pagination-juicy-select-items-per-page" class="kitchensink-juicyselect dropdown kitchensink-test-juicy-select"
+                          value="{{model.EntriesPerPage$::change}}"
+                          options="{{model.PageEntries}}"
+                          text-property="Text"
+                          value-property="Amount">
+            </juicy-select>
+            <p slot="kitchensink/pagination-page-number" class="kitchensink-test-pagination-label">page {{model.CurrentPage}} of {{model.TotalPages}}</p>
```
PatrykSzwer: I would left that pagination label under pagination controls after all. I think It's most intuitive that It's on the bottom part of page.

src/KitchenSink/PaginationPage.json.cs
```diff
@@ -16,143 +17,137 @@ protected override void OnData()
...
-                    if (i > 0 && i < this.TotalPages + 1)
+                    if (i > 0 && i < TotalPages + 1)
                     {
                         CreateButton(i);
                     }
                 }
             }
-            this.DisableLast = this.CurrentPage == this.TotalPages;
-            this.DisableFirst = this.CurrentPage == 1;
+            DisableLast = CurrentPage == TotalPages;
```
PatrykSzwer: Same suggestion with free space as above.

src/KitchenSink/PaginationPage.json.cs
Mackiovello: There is no such guideline from what I know. The code becomes easier to read without it, I think. 

Summoning @warpech for a comment

src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
Mackiovello: On [Amazon](https://www.amazon.com/s/ref=br_pdt_mgLwt?_encoding=UTF8&rh=n%3A172282%2Cn%3A%21493964%2Cn%3A541966&srs=10112675011&pf_rd_m=ATVPDKIKX0DER&pf_rd_s=&pf_rd_r=MMRP3BRDPE8DEFAVQ3Q0&pf_rd_t=36701&pf_rd_p=316d1693-e3b5-4660-bad1-b9586160de85&pf_rd_i=desktop) and many other places, this control is above. 

On the other hand, this is how it looks in the people app, so that's probably what I should do:
pagination

src/KitchenSink/PaginationPage.json.cs
Mackiovello: That's a good suggestion, I'll fix it

src/KitchenSink/PaginationPage.json.cs
Mackiovello: I'm wrong, there such a guidelines [here](https://starcounter.gitbooks.io/guidelines/content/coding_style/csharp_code_formatting.html)

Not sure if it applies to Typed JSON classes or if it even makes a difference

src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
PatrykSzwer: Yes, probably You're right, that we should stay consistent between our apps.

src/KitchenSink/PaginationPage.json.cs
PatrykSzwer: Yes, That's what I was talking about, there is only an entry: 

> We use this. when referring to a current class member

and I'm not sure either if It should be applied to typed JSON classes.

src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
Mackiovello: The layout above is actually not possible since the width is restricted, I think this is the second best option:
newlayout

Mackiovello: As you can see above, the drop down is too short. Previously this was solved with a selector like this:

```css
juicy-select select { width: 200px; }
```

That's not valid BEM. Do you know a good alternative solution?

src/KitchenSink/PaginationPage.json.cs
warpech: It applies especially to Typed JSON classes. Archived discussion here: https://github.com/Starcounter/Guidelines/issues/5

Mackiovello:
Everything should be fixed now. Could you take a look @PatrykSzwer?

PatrykSzwer:
LGTM

Mackiovello:
Thank you! Merging

src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
```diff
@@ -1,15 +1,15 @@
...
     <template is="dom-bind">
         <p slot="kitchensink/html-header" class="lead">Pagination can be implemented in the code-behind by using <code>FETCH</code> to determine how many entries there should be on every page and <code>OFFSET</code> to retrieve the right entries</p>
-        <juicy-select slot="kitchensink/pagination-juicy-select-items-per-page" class="kitchensink-juicyselect dropdown kitchensink-test-juicy-select"
-                        value="{{model.EntriesPerPage$::change}}"
-                        options="{{model.PageEntries}}"
-                        text-property="Text"
-                        value-property="Amount">
-        </juicy-select>
+        <select slot="kitchensink/pagination-juicy-select-items-per-page" 
+                value="{{model.EntriesPerPage$::change}}"
```
warpech: `value=""` binding does not work in `select`. See: https://github.com/StarcounterApps/KitchenSink/issues/126

The reason for having `juicy-select` is that you don't really get preselection of `EntriesPerPage$` without it.

warpech: To really test that I propose to change the initial value of `EntriesPerPage$` to `15`.

Mackiovello: Point taken. Would you say that [this](https://github.com/StarcounterApps/KitchenSink/commit/86d5453dc9885cd67a89cfb76f543f3467cd279e) is a fair fix for that? I.e there is no default value in the JSON-by-example, it is instead set to the first option `OnData`.

warpech: No, because in real systems "entries per page" stays saved in user session. So the server needs to preselect the current value of it. The fair fix is presented on the DropdownPage

Issue #132, opened 2017-05-18
Add a button with an icon inside of it


warpech:
To show the explicit solution for the problem https://github.com/Starcounter/level1/issues/3511 add a new subsection in `ButtonPage.html` that explains the problem and shows a solution with the explicit usage of `pointer-events: none`, preferably added as a CSS BEM class.

Add a test for it (clicking on an element inside the button should trigger the click event on the button).

Fixing this will close https://github.com/Starcounter/level1/issues/3511

warpech:
Done in #134

PR #134, opened 2017-05-22
Added button binding issue example. #132

PatrykSzwer:
#132

### DIFF
--- a/src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
@@ -53,7 +53,7 @@
         </div>
 
         <h2 slot="kitchensink/buttonpage-binding-issue-heading" class="kitchensink-heading-2">Button with children</h2>
-        <p slot="kitchensink/buttonpage-binding-issue-description">Because of the issue in Polymer 1.x, data binding for button with child element cause WebSocket error. To workaround this issue, You can set <code>pointer-events</code> property to <code>none</code> at buttons child element.</p>
+        <p slot="kitchensink/buttonpage-binding-issue-description">Because of an issue in Polymer 1, data bindings for buttons with child elements cause a WebSocket error. As a workaround, set the <code>pointer-events</code> property to <code>none</code> on the children of the button.</p>
         <p slot="kitchensink/buttonpage-binding-issue-click-label"><b>Click to buy a carrot!</b></p>
         <button slot="kitchensink/buttonpage-binding-issue-button" class="btn btn-sm btn-default kitchensink-toggle-button kitchesink-test-buy-carrot-button" value="{{model.BuyCarrotTrigger$::click}}" onmousedown="++this.value"><span class="glyphicon glyphicon-usd kitchensink-buy-carrot"></span></button>
         <span slot="kitchensink/buttonpage-binding-issue-label" class="kitchensink-test-buy-carrot-reaction-label">{{model.OnBuyReaction}}</span>
--- a/src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
@@ -51,6 +51,12 @@
                 <p class="kitchensink-test-self-button-label">Currently Regenerating!</p>
             </template>
         </div>
+
+        <h2 slot="kitchensink/buttonpage-binding-issue-heading" class="kitchensink-heading-2">Button with children</h2>
+        <p slot="kitchensink/buttonpage-binding-issue-description">Because of the issue in Polymer 1.x, data binding for button with child element cause WebSocket error. To workaround this issue, You can set <code>pointer-events</code> property to <code>none</code> at buttons child element.</p>
+        <p slot="kitchensink/buttonpage-binding-issue-click-label"><b>Click to buy a carrot!</b></p>
+        <button slot="kitchensink/buttonpage-binding-issue-button" class="btn btn-sm btn-default kitchensink-toggle-button kitchesink-test-buy-carrot-button" value="{{model.BuyCarrotTrigger$::click}}" onmousedown="++this.value"><span class="glyphicon glyphicon-usd kitchensink-buy-carrot"></span></button>
+        <span slot="kitchensink/buttonpage-binding-issue-label" class="kitchensink-test-buy-carrot-reaction-label">{{model.OnBuyReaction}}</span>
     </template>
     <script>
         (function () {
--- a/src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
@@ -53,7 +53,7 @@
         </div>
 
         <h2 slot="kitchensink/buttonpage-binding-issue-heading" class="kitchensink-heading-2">Button with children</h2>
-        <p slot="kitchensink/buttonpage-binding-issue-description">Because of an issue in Polymer 1, data bindings for buttons with child elements cause a WebSocket error. As a workaround, set the <code>pointer-events</code> property to <code>none</code> on the children of the button.</p>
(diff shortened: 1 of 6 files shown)


PatrykSzwer:
@Mackiovello, could You take a look on presentation of this issue, and WDYT about it, if It's clear enough etc.?

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
```diff
@@ -51,6 +51,12 @@ <h5 slot="kitchensink/buttonpage-self-heading">Self Disabling/Enabling Button</h
                 <p class="kitchensink-test-self-button-label">Currently Regenerating!</p>
             </template>
         </div>
+
+        <h2 slot="kitchensink/buttonpage-binding-issue-heading" class="kitchensink-heading-2">Button with children</h2>
+        <p slot="kitchensink/buttonpage-binding-issue-description">Because of the issue in Polymer 1.x, data binding for button with child element cause WebSocket error. To workaround this issue, You can set <code>pointer-events</code> property to <code>none</code> at buttons child element.</p>
```
Mackiovello: Some small grammar changes:

Because of an issue in Polymer 1, data bindings for buttons with child elements cause a WebSocket error. As a workaround, set the `pointer-events` property to `none` on the children of the button.  

Other than this, I think it's great!

PatrykSzwer: I've fixed it. Tell me if everything else is ok and I'll merge that.

warpech:
@miyconst could you pls review this?

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
Mackiovello: LGTM!

test/KitchenSink.Tests/Test/SectionNumber/ButtonPageTest.cs
```diff
@@ -72,5 +73,15 @@ public void ButtonPage_DisabledButton()
             Assert.IsTrue(WaitForText(_buttonPage.AddCarrotsLabel,
                 "You have 1 imaginary carrots", 5));
         }
+
+        [Test]
+        public void ButtonPage_ButtonWithChildren()
+        {
+            Assert.IsTrue(WaitForText(_buttonPage.BuyCarrotLabel,
+                "You don't have any carrots", 5));
+            _buttonPage.ClickButtonBuyCarrot();
```
warpech: The test is not entirely correct. It should click on the **span** inside the button to prove that it works.

PatrykSzwer: There is a problem about that. Selenium won't let me click span inside button that easy. Using traditional method I'm getting "Other element would receive the click" exception. Another way to do it would be by using Selenium Actions, but in that case, It's not implemented in geckodriver and firefox test is failing. If You know some workaround for that, please let me know and I'll change it.

warpech: Maybe execute it through JavaScript? It is possible via `ExecuteScript`, as used in: https://github.com/StarcounterApps/KitchenSink/blob/e382432b929057feb6850e9823533d086fb9b427/test/KitchenSink.Tests/Ui/BasePage.cs#L96

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
```diff
@@ -51,6 +51,12 @@ <h5 slot="kitchensink/buttonpage-self-heading">Self Disabling/Enabling Button</h
                 <p class="kitchensink-test-self-button-label">Currently Regenerating!</p>
             </template>
         </div>
+
+        <h2 slot="kitchensink/buttonpage-binding-issue-heading" class="kitchensink-heading-2">Button with children</h2>
+        <p slot="kitchensink/buttonpage-binding-issue-description">Because of an issue in Polymer 1, data bindings for buttons with child elements cause a WebSocket error. As a workaround, set the <code>pointer-events</code> property to <code>none</code> on the children of the button.</p>
```
PatrykSzwer: Text is updated, please take a look and tell me if it's ok now 👍

test/KitchenSink.Tests/Test/SectionNumber/ButtonPageTest.cs
PatrykSzwer: I've already tried:
`            IJavaScriptExecutor executor = (IJavaScriptExecutor)Driver;
            executor.ExecuteScript("arguments[0].click();", Driver.FindElement(By.CssSelector(".kitchesink-test-buy-carrot-button")));`
but It's not doing anything. I'll keep trying with this.

PatrykSzwer: FYI:
`            IJavaScriptExecutor executor = (IJavaScriptExecutor)Driver;
            executor.ExecuteScript("document.getElementsByClassName('kitchesink-test-buy-carrot-button')[0].click();");`
also not working.

warpech: 1. `click` is not among events observed by PuppetJs dirty checking.
2. Perhaps Polymer does not know what a `click()` is. Polymer observes events triggered explicitly by `dispatchEvent`.

Try:

```
executor.ExecuteScript("var clickEvent = document.createEvent('MouseEvents'); clickEvent.initEvent("mouseup", true, true); arguments[0].dispatchEvent(clickEvent);");
```

PatrykSzwer: Tried this in a way like that:
```
executor.ExecuteScript("var clickEvent = document.createEvent('MouseEvents'); clickEvent.initEvent('mouseup', true, true); document.getElementsByClassName('kitchesink-test-buy-carrot-button')[0].dispatchEvent(clickEvent);");

still no reaction.

**Comment by warpech** [comment]
Ping @miyconst. Could you pls re-review?

**Comment by warpech** [comment]
Ping @miyconst

**Comment by warpech** [comment]
I would really like to merge when the test tests the right thing (clicking on `span`). @PatrykSzwer do you have a line of code that does this and works in Chrome Dev Tools? Then it should also work in Selenium's `executeScript`.

**Comment by PatrykSzwer** [comment]
The biggest problem is to find the line that does work in Chrome Dev Tools.

**Comment by warpech** [comment]
@alshakero can you pls help with that? @PatrykSzwer will catch you on Slack

**Comment by alshakero** [comment]
According to my investigation, it's not the tests' fault. Go to [http://127.0.0.1:8080/KitchenSink/Button](http://127.0.0.1:8080/KitchenSink/Button)
And execute:
```js
var span = document.querySelector('.kitchesink-test-buy-carrot-button>span')
Object.assign(span.style, { position: 'absolute', top: '1005px', left: '450px'})
```
To take it outside the button, click it, and you'll see the span isn't effectively clickable.

alshakero:
So, I managed to reproduce that Polymer bug, in other words I managed to click the span in a way that satisfies both Polymer and jsonpatch, but the value was set to `undefined`.

Basically I overruled the `pointer-events: none` and went back to square one.

alshakero:
As far as I understand, the Polymer issue arose when the child inside the button is clicked producing a defective event that ruined the value, causing an exception. 

Now that we added `pointer-events: none`, span clicks don't mean anything and the event is originating from the button itself correctly, and it's working without any issue. 

The question is why do we need to handle children clicks to test for it? Unless children are `absolute`-ly outside the button, I can't see another use case. And is it expected for a child that is outside a button to be clicked? And handle that click?

warpech:
Wow, I think that you've nailed it @alshakero. Our solution for the problem https://github.com/Polymer/polymer/issues/2497 is to ignore pointer events on a `<span>`. Then, I asked to test it by performing a `click` event on the `span` - which is a pointer event indeed!

According to https://stackoverflow.com/questions/14176988/why-is-document-elementfrompoint-affected-by-pointer-events-none, this method should work:

```js
var span = document.querySelector(".kitchensink-buy-carrot");
span.scrollIntoView();
var rect = span.getBoundingClientRect();
var hittableElem = document.elementFromPoint(rect.left + 1, rect.top + 1);
var mousedown = document.createEvent('MouseEvents');
mousedown.initEvent('mousedown', true, true);
hittableElem.dispatchEvent(mousedown);
var click = document.createEvent('MouseEvents');
click.initEvent('click', true, true)
hittableElem.dispatchEvent(click);
```

PatrykSzwer:
Surprising as it is - It's working! I've pushed test with script inside.

Issue #135, opened 2017-05-31
Add a page that mixes the view-model and the local state


warpech:
Some users need to keep local state of objects, for example:

- have an array of items
- present each item in a `dom-repeat`
- for each item have a local property that represents the mouse hover. 
  - That local property should be available to Polymer but not part of the view-model that is synced with the server

@alshakero could you pls prepare a page for that in KitchenSink?

alshakero:
I think this is closable with https://github.com/StarcounterApps/KitchenSink/pull/136 now.

warpech:
Closing. Thanks @alshakero!

PR #136, opened 2017-05-31
Add LocalViewModel page to demonstrate local (non-synced) state example

alshakero:
It seemed weirdly simple to me.

_I apologise if I misunderstood something_ 🙇

Implements: https://github.com/StarcounterApps/KitchenSink/issues/135


src/KitchenSink/DropdownPage.json.cs
```diff
@@ -46,7 +46,7 @@ protected override void OnData()
 
         public string SelectedProductKeyBind
         {
-            get => SelectedProduct?.Key;
+            get { return SelectedProduct == null ? null : SelectedProduct.Key; }
```
warpech: Why this change? Is the app not building for you?

This is a C# 7.0 feature and KitchenSink is configured as a .Net 4.5 project. Maybe that is the problem. Should we change KitchenSink to be a .Net 4.7 project? @joozek78, @ubbeK please advice

src/KitchenSink/app.config
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5"/></startup></configuration>
```
warpech: @joozek78, @ubbeK: any idea why VS creates this file for @alshakero?

@alshakero what is your Windows and VS version?

src/KitchenSink/wwwroot/KitchenSink/LocalViewModelPage.html
```diff
@@ -0,0 +1,44 @@
+﻿<template>
+    <p slot="kitchensink/localviewmodel-description" class="lead">This page demonstrates the ability to keep parts of your state local (unsynced with the server).</p>
+    <template is="dom-bind">
+        <p><em>Please hover an item to see it in action</em></p>
+        <ul>
+            <template is="dom-repeat" items="{{model.People}}">
+                <person-item model="{{item}}" />
+            </template>
```
warpech: This is obvious, but @malx122 has a little bit more explicit HTML in few cases. Do you know how to make the following work?

```html
<ul>
  <template is="dom-repeat" items="{{model.People}}">
  <li>
    <hover-observer is-hovered="{{xxx}}">
         <span>{{item.Name}}</span>
         <template if="{{xxx}}">
            <span>Hovered</span>
          </template>
    </hover-observer>
  </li>
  </template>
</ul>
```

src/KitchenSink/DropdownPage.json.cs
alshakero: Yes, it wasn't building.

src/KitchenSink/app.config
alshakero: Windows 10 Pro
Version 1703 (OS Build 15063.332)

Microsoft Visual Studio Community 2015
Version 14.0.25431.01 Update 3

Microsoft .NET Framework
Version 4.7.02046

src/KitchenSink/DropdownPage.json.cs
warpech: While we work on how to better inform the developer that C# 7.0 is required, could you @alshakero pls revert the above line and make sure you have C# 7.0 installed?

alshakero: @warpech of course. I'll look into it.

joozek78: @alshakero to support C#7 please install VS 2017. It's as free as 2015 and switching over shouldn't be much of a hassle

alshakero:
@warpech I tried every possible scenario and `xxx` in your example will be always overwritten be the next iteration. `dom-repeat` offers only `item` as a loop-independent binding and I couldn't overcome that.

And ignoring mechanism seems necessary. I suggest waiting for Palindrom 3.0.0 and then showing the `_isProxified` example.

warpech:
Have you considered implementing `{{xxx}}` as `{{getItemMeta(item).IsHovered}}` where `getItemMeta` is a template method implemented using WeakMap? 

@tomalec, @miyconst any other ideas? Context: #135

alshakero:
@warpech I created computed properties that depend on WeakMaps. But it didn't occur to me to use a function inside the binding. Will give it a shot now.

alshakero:
@warpech apparently Polymer doesn't parse `{{{func().value}}}` pattern, it must be a function only. And I guess a separate function for reach property isn't nice.

However the `Object.defineProperty` idea worked and it's clean enough 😍

alshakero:
Please only pay attention to the HTML file. I had some trouble with VS 2017 and proceeded with VS 2015. Once we agree on a way I'll clean up.

tomalec:
For Polymer 2 they [claim](https://www.polymer-project.org/2.0/docs/devguide/templates#dom-repeat) that
> For each instance, it creates a new data binding scope 

So hopefully just `"{{xxx}}"` should work.


As we theoretically [agreed on stopping Polymer 1.x development](https://github.com/Starcounter/RebelsLounge/issues/66) I would suggest not to waste to much time on it.

warpech:
Please forgive my two clumsy commits. I have played with this a bit. It is hard to propose something significantly better, though I managed to add separation of concern: `define-local`, `hover-observer`. 

WDYT?

> As we theoretically [agreed on stopping Polymer 1.x development](https://github.com/Starcounter/RebelsLounge/issues/66) I would suggest not to waste to much time on it.

I wouldn't say its a waste of time. Palindrom 3.0.0 removed `ignoreAdd` and we must propose a substitute for it.

alshakero:
> For Polymer 2 they claim that

And they claim [exactly the same](https://www.polymer-project.org/1.0/docs/devguide/templates#dom-repeat) in Polymer 1.0, so I guess we shouldn't hope for the good. 

I think @warpech's solution is as clean as it can get. I really like it. Would you like me to clean up the branch and resolve conflicts? And don't worry at all about the commits, I'll squash commits on this branch to a nice single commit when we finish.

warpech:
Please do. If there are better ideas later, we can always improve. Also, could you pls rename the page "Local view model" to "Client Local State"? Can you pls try to add a test for it?

alshakero:
@warpech as for the tests. It seems like the tests follow the old grouping of pages. Is someone going to fix that? If no I can try and add tests on top of the current state.

src/KitchenSink/wwwroot/KitchenSink/ClientLocalStatePage.html
```diff
@@ -0,0 +1,70 @@
+﻿<template>
+    <p slot="kitchensink/localviewmodel-description" class="lead">This page demonstrates the ability to keep parts of your state local (unsynced with the server).</p>
```
warpech: Please remove `class="lead"` and add a proper header to the page. Similarly to how other pages were refactored recently on the `develop` branch

warpech:
> @warpech as for the tests. It seems like the tests follow the old grouping of pages. Is someone going to fix that? If not I can try and add tests on top of the current state.

Good catch. Please add the tests on top of the current state for now.

alshakero:
@warpech Thanks a lot! Merging.

PR #149, opened 2017-06-29
Fix missing property from ButtonPage.json

alshakero:
This PR is just to document a fix for https://github.com/Palindrom/palindrom-polymer-client/issues/42

### DIFF
--- a/src/KitchenSink/ButtonPage.json
+++ b/src/KitchenSink/ButtonPage.json
@@ -12,5 +12,6 @@
   "CarrotsReaction": "You don't have any carrots",
   "CarrotEngineReaction": "Carrot engine is off",
   "OneCarrotReaction": "You don't have any carrots",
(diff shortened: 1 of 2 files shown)


warpech:
Great find, but I think that this is not a correct fix.

Correct would be to change: 

```
template.set("model.AddCarrots$", ++template.model.AddCarrotsTrigger$);
```

to 

```
template.set("model.AddCarrotsTrigger$", ++template.model.AddCarrotsTrigger$);
```

The problem was introduced in https://github.com/StarcounterApps/KitchenSink/pull/108. My bad for not seeing it in the review.

alshakero:
@warpech wow thank you. This is great news! I was worried that it was a common practice to add properties to root. And that removing `ignoreAdd` would have had caused a lot of trouble! What a phew moment.

warpech:
@Mackiovello could you pls review? Maybe other HTML needs fixing too?

PR #151, opened 2017-06-30
Update bower components to current versions 

warpech:
Update bower components to current versions, including hybrid paper elements.

Part of preparation for https://github.com/Starcounter/RebelsLounge/issues/72

Work in progress. Left to do:

1. fix few tests (file upload, date picker)
2. write what components still use Shadow DOM v0


warpech:
Fixed failing tests, but the `<content>` tag is still to be found in few elements.

PR #155, opened 2017-07-10
update bower components to current versions

warpech:
To clarify: the purpose of this PR is to upgrade custom elements used in KitchenSink to newer versions that use `<slot>` elements instead of `<content>` and are compatible with Shadow DOM v1 and compatible with Polymer 1.9 - 2.0 (so called hybrid elements).

It is part of preparation for Starcounter/RebelsLounge#72

Changes made so far by me:

- updated `paper-` elements
- for starcounter-upload, now a auto-upload attribute is required to pass tests
- for paper-input, CSS selectors needed to be changed

Searching for `<content>` in this branch shows that still the following custom elements need to be updated:

- google-map (take newer version from GH)
- hot-table (https://github.com/handsontable/hot-table/issues/44)
- juicy-select (ours, need upgrading. Tracked in https://github.com/Juicy/juicy-select/issues/6)
- pikaday-decorator (newer version is problematic, consider replacing with vaadin-date-picker)
- x-breadcrumb (ours, need upgrading. Tracked in https://github.com/Juicy/x-breadcrumb/issues/28)

I can take `hot-table`, @Mihaiii can you pls take the other ones?


warpech:
To clarify: the purpose of this PR is to upgrade custom elements used in KitchenSink to newer versions that use `<slot>` elements instead of `<content>` and are compatible with Shadow DOM v1 and compatible with Polymer 1.9 - 2.0 (so called hybrid elements).

Mihaiii:
Tried to replace "pickaday" with "vaadin-date-picker" (no version provided).
It tried to install "vaadin-date-picker" v1.2.3, but that didn't worked because it requires `paper-input#^1.1.5`, which contradicts "paper-input#^2.0.1" that we now have in `bower.json`.
So I tried to install lates version ("v2.0.0-beta2"), but there were problems with the version of "iron-resizable-behavior".
*"Chart-elements" has "Iron-resizable-behavior#~1.0.2"*, "google-map"  has "iron-resizable-behavior#^1.0.5" and "iron-collapse" has "iron-resizable-behavior#^1.0.0".
"Google-map" is already at the latest version ("1.2.0").
Because the specification said "google-map (take newer version from GH)", I change it in bower.json (`"google-map": "GoogleWebComponents/google-map#2.0-preview"`) and ran `bower update`.
"Iron-collapse" was removed because it's not used.
"Chart-elements" is already at the latest version (v3.1.3). There is a "polymer-2.0" branch, so I took it (by using it in `bower.json`) and after that I ran `bower update`.
"http://localhost:8080/KitchenSink/Chart" doesn't work after the update of "chart-elements" so I had to revert everything.
Please consider replacing `chart-elements`.

warpech:
I think that at least some of the elements could be removed, but as agreed yesterday on https://github.com/Starcounter/Minutes/pull/64, we should not worry about Polymer elements right now, only about `<content>` and `::content` and `/deep/` in our views and plain JavaScript custom elements.

Thanks @Mihaiii for your investigation, we will repeat that few weeks from now :)

I think that PR is ready for merging then. @PatrykSzwer do you agree?

warpech:
Thanks, merging.

PR #158, opened 2017-07-19
Add declarative shadow dom

warpech:
Changes are explained in the commit messages.

This should be merged to KitchenSink `develop` after the branch https://github.com/Starcounter/level1/tree/starcounter-include-CEv1 is merged to Starcounter `v2.3`. Otherwise KitchenSink does not render pages.

Last commit says "WIP" but actually it seems to be final change that is needed regarding `declarative-shadow-dom`.


warpech:
@PatrykSzwer is on vacation. Can you review @Mihaiii?

PR #177, opened 2017-10-02
Fixed Nested childs using Self.GET way

### LINKED ISSUE #167: Can't create more than one nested child
### Steps to reproduce
1. Run https://github.com/StarcounterApps/KitchenSink/commit/d42dba70007a1be2fe7987de6460910967232e3c (current latest version on develop)
2. Go to `http://localhost:8080/KitchenSink/Nested`
3. Click "Add child"
3. Within the child, click again "Add child"

### Expected result

nested1

### Actual result

nested2

### Technical information
It doesn't work starting with SC 2.3.1.7034 (22 aug. 2017). So it works with SC 2.3.1.7018 (18 august 2017). Couldn't find any version in between those two.

@tomalec : I'm not sure if you're the right person to tag here, but could you please have a look if you are ?

un-tone:
Fixed #167

### DIFF
--- a/src/KitchenSink/NestedPartial.json.cs
+++ b/src/KitchenSink/NestedPartial.json.cs
@@ -1,5 +1,5 @@
-using Starcounter;
 using System;
+using Starcounter;
 
 namespace KitchenSink
 {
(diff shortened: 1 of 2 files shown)


un-tone:
There is a lot of failing tests. I created a separate issue for it - #178

un-tone:
I suppose this way with `Self.GET` is more preferable in this case, right?

warpech:
This is the advised way. Please merge. Thanks for the investigation!

un-tone:
Added the new test. Please check

warpech:
Two flaky tests are still failing on TC. I know this is unrelated to this PR. I created a new issue for that: https://github.com/StarcounterApps/KitchenSink/issues/181

PR #184, opened 2017-10-09
Update handsontable/hot-table@0.6.1 - SDv1 compatible

tomalec:
Teamcity build http://teamcity.starcounter.org/viewType.html?buildTypeId=Apps_KitchenSink_Develop&branch_Apps_KitchenSink=hot-custom-element-upgrade&tab=buildTypeStatusDiv


tomalec:
Those two tests that are failing are flaky
https://github.com/StarcounterApps/KitchenSink/issues/181

warpech:
This works well. The right column fails to update in Chrome but this is not a regression. I will make another issue about that.

PR #187, opened 2017-10-09
Problem: Datagrid does not work well in Chrome (#186)

### LINKED ISSUE #186: Datagrid page needs improvements
The datagrid page should get the following improvements:

- add a paragraph that explains how the right column gets updated by the server when you type a name of an animal
- in Chrome, the right column does not get automatically refreshed. That's a bug, fix that.
- add a test for the above feature

Assigning this on me.

warpech:
Solution: 

- Add description of the intended behavior in `DatagridPage.html`
- Temporarily add a observer function in `DatagridPage.html`
- Create an issue in hot-table
- Add tests

Fixes #186


warpech:
Great, thanks for the review!

PR #188, opened 2017-10-10
Update pikaday-decorator@0.9.0(latest) - Shadow DOM v1 ready

tomalec:
Fixes: https://github.com/StarcounterApps/KitchenSink/issues/182
https://github.com/Starcounter/RebelsLounge/issues/72


warpech:
Looks great! Pls merge

alshakero:
If I'm welcome to comment, I think using `latest` isn't a good idea. If they released a major with breaking changes our app would break. I'd lock it to a major. Releasing jsonpatch 2.0 with breaking changes caused some issues for us because maybe deps had it as `>=1.0.0`.

warpech:
Actually are these two lines needed? https://github.com/StarcounterApps/KitchenSink/blob/84cec5cbc302db6ca78f162e23f05663b6f19600/src/KitchenSink/bower.json#L33-L34

Moment and Pikaday are already dependencies of Pikaday-decorator, there is no need in repeating them here.

Issue #181, opened 2017-10-06
Two flaky tests failing often on TC


warpech:
These two tests fail in 50% of time on TC:

```
AutoCompletePageTest(Firefox).AutoCompletePage_FillCountryNameThenSelectCountry
RedirectPageTest(Firefox).RedirectPage_ClickingOnFruitShouldChangeUrlAndText
```

Sometimes this fails:

```
CheckboxPageTest(Firefox).CheckboxPage_CheckboxUncheckedAndCheckedAgain
```

@un-tone do you have any idea why this might be? @PatrykSzwer was investigating this but could not find the reason: https://github.com/StarcounterApps/KitchenSink/pull/174#issuecomment-334487543

un-tone:
There was 2 running `geckodriver.exe` processes in background. I guess it was a reason.

I also updated selenium binary and drivers to last versions on ScBuildAppWin01.

warpech:
Thanks, but I think that was not it. A new build failed yesterday: https://teamcity.starcounter.org/viewLog.html?buildId=33879&buildTypeId=Apps_KitchenSink_Develop

Btw. Could you please update `CHANGELOG.txt` on the desktop of the build agent?

warpech:
Fixed in #189

un-tone:
> Could you please update CHANGELOG.txt on the desktop of the build agent?

Done.

PR #189, opened 2017-10-13
Fix teamcity tests

warpech:
Finally, I think I have fixed the TeamCity flaky tests (#181).


warpech:
I will squash and merge this. What can go wrong?

warpech:
@un-tone and @PatrykSzwer still, pls review this when back from vacation.

un-tone:
So the solution in [this line](https://github.com/Starcounter/KitchenSink/pull/189/files#diff-4556d20ff9b33a014a25690691f407c7R54):
```csharp
Driver.Manage().Window.Maximize();
```
👍 Good to know about such unobvious things which can break tests.

PatrykSzwer:
There is actually one thing that makes me suspicious. Tests were running in maximized windows already. See here: https://github.com/Starcounter/KitchenSink/blob/develop/test/KitchenSink.Tests/Utilities/WebDriverManager.cs#L36

But since this actually fixed our issue maybe there was something wrong in using this function there. I thing we should remove this line and check if It will break something.

warpech:
Wow, that's interesting!

https://github.com/Starcounter/KitchenSink/blob/831c02c6a20f6a404b4a9e0419c296690df2c840/test/KitchenSink.Tests/Utilities/WebDriverManager.cs#L36 seems to really not work in Firefox, because the screenshots made in falling tests by Selenium were 100x200 pixels, not more. 

Maybe this is related to https://github.com/mozilla/geckodriver/issues/900 or one of the issues linked to it.

warpech:
The bottom line is that in FF 55, `Maximize()` toggles the window size instead of maximizing it. Maybe we should stop using `Maximize()` and use this instead:

```
Driver.Manage().Window.SetPosition(new Point(0,0));
Driver.Manage().Window.SetSize(new Dimension(1024,768));
```

PatrykSzwer:
Maybe way of implementing this matters for gecko. In line that I referenced Maximize is used on instance like this:
`IWebDriver eventDriver = new KitchenSinkTestEventListener(driver);`
When, In Your solution, maximize is called directly on Driver class. 
I'm not sure if this really makes a difference, just guessing, according to that previous solution didn't work and this does.

warpech:
Let's continue the discussion in https://github.com/Starcounter/KitchenSink/issues/190

Issue #190, opened 2017-10-16
Tests problem with Window.Maximize()


warpech:
After my last changes in https://github.com/Starcounter/KitchenSink/pull/189, KitchenSink tests pass, but have rather ridiculous double call to `Window.Maximize()`:

https://github.com/Starcounter/KitchenSink/blob/607cfcdd9cd66d2fa90481db2f186572aeda50b9/test/KitchenSink.Tests/Utilities/WebDriverManager.cs#L36

https://github.com/Starcounter/KitchenSink/blob/607cfcdd9cd66d2fa90481db2f186572aeda50b9/test/KitchenSink.Tests/Test/BaseTest.cs#L54

The problem that I fixed was connected to the fact that Firefox 55 does not correctly maximize a window: https://github.com/mozilla/geckodriver/issues/900

We need to find a solution that does not involve calling `Maximize()` twice.

In order to have rather predictable tests, maybe we should set a specific window size, instead of relying on maximizing it?

```
Driver.Manage().Window.SetPosition(new Point(0,0));
Driver.Manage().Window.SetSize(new Dimension(1024,768));
```

PatrykSzwer:
I've actually removed this one, unnecessary Maximize call, and locally, everything seems working. I've created PR #191  for this, We will see if tests on teamcity will fail or not.

PatrykSzwer:
I've replaced Manage().Window calls with options argument. I think It's most elegant and most of all reliable solution.

PatrykSzwer:
Closing as done.

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/KitchenSink/releases/tag/2.0.0
- published in the App Warehouse
- blogged at https://starcounter.io/kitchensink-2-0-0-released-includes-34-common-ui-patterns/

PR #191, opened 2017-10-16
Removed unnecessary maximize call

PatrykSzwer:
for #190

### DIFF
--- a/test/KitchenSink.Tests/Test/BaseTest.cs
+++ b/test/KitchenSink.Tests/Test/BaseTest.cs
@@ -50,8 +50,6 @@ namespace KitchenSink.Tests.Test
             {
                 Assert.Ignore(Config.BrowserDictionary[_browser] + " is on browsers ignore list");
             }
-
-            Driver.Manage().Window.Maximize();
         }
 
         [OneTimeTearDown]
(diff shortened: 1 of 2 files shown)


PatrykSzwer:
I've run tests 4 times and seems like my last change fixed our problem. Please check by yourself.

PatrykSzwer:
I've came to some interesting results.
Seems like, what is failing our tests in firefox is... maximizing itself! When I changed way of maximizing to argument in options, I though everything is working, but turns out that Firefox doesn't have argument for maximizing. What really happen is that firefox had maximizing turned off completely. And THAT fixed tests. 
Now there is a question. Should we completely shut down maximization in firefox? Seems like It's not needed for tests to be run properly.

warpech:
If maximization is not needed, then why remove it only from Firefox? It should be consistently removed from all drivers.

PatrykSzwer:
Please check if tests passes locally for you, because I have some flacky ones, but on teamcity everything seems fine.

warpech:
Tests locally pass but I get a lot of TearDown errors:



Do you also get these?

PatrykSzwer:
Yes I've started getting those also couple days ago. I though It can be related to some update or something, even Resharper is also showing them with red color but not categorizing as failed tests. But I've just looked into it further and found out that It's caused by getting snaphsot in teardown method. I'll fix it right away.

PR #192, opened 2017-10-17
Remove obsoleted

Mackiovello:
Small update to remove the obsoleted `Session.Data` and `Partial` class.

`json.Session` should also be removed, but I wasn't able to make it work.


Mackiovello:
There we go, everything you mentioned is fixed. I'm going to look into HandleFile.cs and refactor a bit since the code there is messy (four `out` parameters etc.). I'll do that in another PR

PR #194, opened 2017-10-18
Refactor

Mackiovello:
Broke up a large class into more manageable pieces. I still don't like the way `UploadTask` handles its state, but it's not worth fixing right now.


Mackiovello:
It seems like this broke something. Eh... let's not bother then. Closing.

Issue #10, opened 2015-12-09
The URL to the map can't be /KitchenSink/map


warpech:
I would like to change the URL to the page about `<google-map>` from:

```
http://localhost:8080/KitchenSink/Geo
```

to:

```
http://localhost:8080/KitchenSink/Map
```

but then it stops working:

<img width="1431" alt="screen shot 2015-12-09 at 10 45 56" src="https://cloud.githubusercontent.com/assets/566463/11681909/0ee3788a-9e62-11e5-86ef-9d831626533c.png">

Results in exactly the same problem as:

https://github.com/Starcounter/Starcounter/issues/3126

aloscha:
Indeed. The problem was with naming url MAP. After changes in https://github.com/StarcounterSamples/KitchenSink/tree/Map_Link_%2310 to GEO, everything works like a charm

warpech:
Merged, thanks!

warpech:
But this doesn't solve the orignal problem that I want the url to be `.../map`, not some workaround word. I have pushed the code that reproduces the problem to a new branch: https://github.com/StarcounterSamples/KitchenSink/tree/Starcounter_issue_3126

Steps to reproduce:

1. Build and start app 
2. Go to `http://localhost:8080/KitchenSink` in the web browser
3. Click on the "Map" link

Expected: 

Should morph to the desired URL (`http://localhost:8080/KitchenSink/Map`)

Actual:

Throws error in the Chrome console, because wrong JSON response is returned from the server.

PatrykSzwer:
'/kitchensink/map` is working as expected now. Closing as fixed.

warpech:
There is no page with URI `/kitchensink/map`.

We have `/kitchensink/geo` for the reason that `/kitchensink/map` did not work.

Could you please change the URI to `/kitchensink/map` and check if it works?

PatrykSzwer:
Yes It's working. 
I'll create PR for that change.

PR #204, opened 2017-11-27
Changed uri /Geo to /Map

PatrykSzwer:
For #10 

I've changed only Uri associated with maps, not file names etc. Tell me if You want me to change them as well, to get rid of any `Geo` word usages.

### DIFF
--- a/src/KitchenSink/Program.cs
+++ b/src/KitchenSink/Program.cs
@@ -53,7 +53,7 @@ namespace KitchenSink
             Handle.GET("/KitchenSink/partial/integer", () => new IntegerPage());
             Handle.GET("/KitchenSink/integer", () => WrapPage<IntegerPage>("/KitchenSink/partial/integer"));
 
-            Handle.GET("/KitchenSink/partial/Geo", () =>
+            Handle.GET("/KitchenSink/partial/Map", () =>
             {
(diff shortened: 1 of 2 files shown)


warpech:
> I've changed only Uri associated with maps, not file names etc. Tell me if You want me to change them as well, to get rid of any `Geo` word usages. 

Yes, please :)

warpech:
Then this means we cannot do this PR because the problem from https://github.com/Starcounter/KitchenSink/issues/10 still exists.

PatrykSzwer:
I've reverted my workaround and reported this as an issue: https://github.com/Starcounter/Home/issues/327

warpech:
Blocked by https://github.com/Starcounter/Home/issues/327

PatrykSzwer:
I talked to @alemoi and he told me that issue reported in Starcounter/Home#327 will not be addressed soon. I guess the best and quickest way will be to rename GeoPage to gMapPage or something similar.

warpech:
I don't think that we have to change a workaround to another workaround. Better to close this PR and wait for the solution in https://github.com/Starcounter/Home/issues/327.

I pasted an explicit link to a commit in https://github.com/Starcounter/Home/issues/327, so now this PR can be closed and the branch can be deleted.

PR #208, opened 2017-12-04
Fix loading a stylesheet on root of an HTML Import

warpech:
Problem: Loading a stylesheet on root of an HTML Import to display a warning (https://github.com/Starcounter/RebelsLounge/issues/222)

Solution: move loading of the stylesheet to the stamped template

### DIFF
--- a/src/KitchenSink/wwwroot/KitchenSink/MasterPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/MasterPage.html
@@ -1,5 +1,5 @@
-﻿<link rel="stylesheet" href="/KitchenSink/style.css">
-<template>
+﻿<template>
+    <link rel="stylesheet" href="/KitchenSink/style.css">
     <template id="root" is="dom-bind">
         <starcounter-include slot="kitchensink/nav" partial="{{model.NavPage}}"></starcounter-include>
         <starcounter-include slot="kitchensink/current" partial="{{model.CurrentPage}}"></starcounter-include>
--- a/src/KitchenSink/wwwroot/KitchenSink/MasterPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/MasterPage.html
@@ -1,5 +1,13 @@
-﻿<template>
-    <link rel="stylesheet" href="/KitchenSink/style.css">
+﻿<link rel="stylesheet" href="/KitchenSink/style.css">
+<script>
+(function(){
+    var importDoc = document.currentScript.ownerDocument;
+    var elem = importDoc.querySelector("link[rel='stylesheet']")
+    document.head.appendChild(elem);        
+})();
+</script>
+
+<template>
     <template id="root" is="dom-bind">
         <starcounter-include slot="kitchensink/nav" partial="{{model.NavPage}}"></starcounter-include>
         <starcounter-include slot="kitchensink/current" partial="{{model.CurrentPage}}"></starcounter-include>


warpech:
I will make a change, please don't review yet.

warpech:
Please review now

tomalec:
Why don't we just move style to the template, to stamp it to the main doc?

warpech:
> Why don't we just move style to the template, to stamp it to the main `document`?

I did that in my first commit, but later I thought that your solution is superior because it is guaranteed to happen once.

tomalec:
We need to choose between:
1. Happens once and is detached from partial actually being stamped.
   When you restamp, or even remove `MasterPage.html`, `<style>` will still be there. 
2. Happens per stamp but it's closely attached to partial.
   When you remove `MasterPage.html` it will be removed as well. When you replace it with another page, the styles will not affect it. When (for some reason) you will stamp it multiple times, you will end up with multiple `<style>`s. 

    So there will be an overhead of duplicated elements, the problem of performance relies on how the browser handles duplicates in such case. Probably it will not fetch it again (we could try `@import` if it does), but for sure it will process it again, to apply correct cascade order. Like for:
   ```html
   <!-- MasterPage.html -->
   <style>
     .foo{
       color: green;
     }
   </style>
   <!-- OtherPage.html -->
   <style>
     .foo{
       color: red;
     }
   </style>
   <!-- MasterPage.html -->
   <style>
     .foo{
       color: green;
     }
   </style>
   ```


So we need to choose which behavior we would actually like.

tomalec:
(previously the behavior was exactly 1.)

warpech:
TL;DR the choice is between:

1. leaky stykes
2. unleaky styles 

Then I think the desired solutions is obviousluy 2. Who wants leaky styles?

> So there will be an overhead of duplicated elements, the problem of performance relies on how the browser handles duplicates in such case. Probably it will not fetch it again (we could try `@import` if it does), but for sure it will process it again, to apply correct cascade order. 

I did a [quick test in JSFiddle](https://jsfiddle.net/Lsswh1ep/1/). Funny enough, Chrome and Firefox dedupe `link` and `@import`. Edge does not dedupe any of those, and in fact makes a 3rd call to `@import` for some reason.

image

-----

The above shows the unfortunately we must go with solution 1. Or is my test wrong?

warpech:
Actually, checking with Fiddler 4 shows me something opposite. Chrome and Edge correctly dedupe both `link` and `@import`, while Firefox does not dedupe `link` (even though it looks deduped in Firefox DevTools).

alshakero:
We can add a very simple ServiceWorker that will cache all HTTP requests on every browser.

tomalec:
What about
3. (or 2.1) Dedup link manually:
```html
<script>
<!-- Custom element to be shared -->
customElements.define('deduped-link', class extends HTMLElement{
    connectedCallback(){
        const href = this.getAttribute('href');
        const tree = this.getRootNode();
        let link = tree.querySelector(`link[rel=stylesheet][href=href]`);
        if(!link){
            link = document.createElement('link');
            link.rel = 'stylesheet';
            link.href = href;
            this.link = link;
            (tree.head || tree).appendChild(link);
        } 
        if(!link.__dedupedScripts){
            link.__dedupedScripts = new WeakMap();
        }
        link.__dedupedScripts.set(this, true);
    }
    disconnectedCallback(){
        const link = thi.link;
        link.__dedupedScripts.delete(this);
        if(link.__dedupedScripts.length === 0){
            link.remove();
        }
    }
});
</script>

<!-- partial html code -->
<template>
    <deduped-link rel="stylesheet" href="/KitchenSink/style.css"></deduped-link>
</template>
```

warpech:
Actually the [CSS3 spec](
https://drafts.csswg.org/css-cascade-3/#at-ruledef-import) says explicitly that `@import` is not deduped. I believe it is the same for `link`:

> When the same style sheet is imported or linked to a document in multiple places, user agents must process (or act as though they do) each link as though the link were to an independent style sheet.

I agree that a custom element is a way to go. 

@tomalec can you make a more generic version that uses `this.attributes.forEach()` in `connectedCallback`? That would work with any links, not just `link rel="stylesheet"`. We will include it in Starcounter.

tomalec:
> @tomalec can you make a more generic version that uses `this.attributes.forEach()` in `connectedCallback`? That would work with any links, not just `link rel="stylesheet"`. We will include it in Starcounter

Wow, that could be asking for troubles in my opinion. Would you like to do the same for `link rel="imports"` for example? what if custom element definition relies on the imported document? then if we will remove it, but the element will be already registered (and in use) it will throw hard to debug errors. (I assume one may use custom element without attaching definition either tofor progressive enhancement  or by assuming that according to spec CE once defined remains defined).

also rewriting all attributes involves a problems like below.
Consider a main document to which we load three views
**main.html**:
```
<link rel="stylesheet" id="foo" title="Main style" href="M.css"/>
```
Then you load
**Aview.html**:
```
<template>
  <deduped-link rel="stylesheet" id="bar" disabled href="M.css"/>
</template>
```
**Bview.html**:
```
<template>
  <deduped-link rel="stylesheet" title="Fancy" preload href="M.css"/>
</template>
```
**Bview.html**:
```
<template>
  <deduped-link rel="prefetch" href="M.css"/>
</template>
```
What you would like to see at the end? Which id should be applied to the element? Or would you create a new instance if there is an id? Which `title` would you like to apply (even if the id+href is matching)? Should `prefetch` be ignored?

Then if you a partial is removed, would you like to restore the attributes to the previous state?

----

I think that's overcomplicating.

Let's have an element that does one thing, serves one purpose and do it correctly and fast.


Alternatively, we could have an element that dedupes the element only if all attrs are matching.

warpech:
> Alternatively, we could have an element that dedupes the element only if all attrs are matching.

This :)

tomalec:
Do you like me to take over this issue? I'd add this element to `Starcounter.ClientFiles`

warpech:
Yes please

alshakero:
> Do you like me to take over this issue? I'd add this element to `Starcounter.ClientFiles`

@tomalec Could you wait for my ping before going for this? `StarcounterClientFiles` will look very different tomorrow and re-basing can be terrible.

warpech:
Changes in `StarcounterClientFiles` are done.

warpech:
Closing in favor of https://github.com/Starcounter/RebelsLounge/issues/234

PR #209, opened 2017-12-05
No-underwear-css class

PatrykSzwer:
For https://github.com/Starcounter/RebelsLounge/issues/75#issuecomment-348955423 step 3.4


warpech:
@PatrykSzwer just showed me a problem with this approach. 

In CSS, `button:not(.no-underwear-css)` has the same specifity as `button.no-underwear-css`.

So, if KitchenSink tries to overwrite Underwear's style of a button using:

```css
button {
  color: red;
}
```

or

```css
.kitchensink-my-button {
  color: red;
}
```

it will fail to overwrite it in both cases.

PR #211, opened 2017-12-07
FlashMessagePage button crashes app. Fix #210

### LINKED ISSUE #210: Triggering messages at Flash Message page crashes application
### Steps to reproduce
1. Go to `/KitchenSink/FlashMessage`.
2. Click on `Trigger message from server side`.
3. Wait couple of seconds.
### Expected result
Message shows up and nothing more happen.
### Actual result
Message is shown but entire app crashes after couple of seconds.

##### Browsers affected
- [x] Chrome

BTW: There is no tests that covers Flash Message view. We should add them to prevent situations like that in the future.

joozek78:
Instead of manual meddling in scheduling and threads,
Joozek78.Starcounter.Async library has been used to
easily implement delay in FlashMessagePage. It required
the project to target framework 4.7 and inlcude weaver.ignore

### DIFF
--- a/src/KitchenSink/FlashMessagePage.json.cs
+++ b/src/KitchenSink/FlashMessagePage.json.cs
@@ -1,3 +1,6 @@
+using System;
+using System.Threading.Tasks;
+using Joozek78.Star.Async;
 using Starcounter;
 
 namespace KitchenSink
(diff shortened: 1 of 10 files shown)


PatrykSzwer:
Please update text on FlashMessagePage about usage of your library.

PR #217, opened 2017-12-20
Polymer 2 hybrid views

warpech:
Make the views run with Polymer 1 or 2.

Requires https://github.com/Starcounter/StarcounterClientFiles/pull/16, so I would wait with merging until  https://github.com/Starcounter/StarcounterClientFiles/pull/16 is released as a version of `StarcounterClientFiles` and shipped in Starcounter 2.3.1

In other words, please review but I will merge only when the build passes with Starcounter 2.3.1

### DIFF
--- a/src/KitchenSink/wwwroot/KitchenSink/AppIcon.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/AppIcon.html
@@ -1,4 +1,4 @@
-﻿<template>
+<template>
     <img slot="kitchensink/kitchensink-app-icon-image" src="/KitchenSink/images/KitchenSink.svg">
     <template is="declarative-shadow-dom">
         <slot name="kitchensink/kitchensink-app-icon-image"></slot>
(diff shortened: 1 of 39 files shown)


warpech:
Thanks! Now that's 110% satisfaction :)

warpech:
The test with 2.3 feature branch passed. That feature branch is now merged to 2.3 (https://github.com/Starcounter/level1/pull/4420), so after the [build](https://teamcity.starcounter.org/viewLog.html?buildId=41252&tab=buildResultsDiv&buildTypeId=Starcounter_V23NightlyWindows) finishes, I can merge this PR.

warpech:
Both checks now green with 2.3 and 2.4! Merging with satisfaction

alshakero:
I learned from this PR many attribute names that should be empty `restamp`, `multiple` etc and now the ViewTranslator knows about them.

See https://github.com/Starcounter/ViewTranslator/commit/b97a55c20a46d90501b5e64c1d6e8399052525c1

Issue #219, opened 2018-01-03
Paginated table does not have explicit slot, is it by desired?


tomalec:
As you can see at https://github.com/Starcounter/KitchenSink/blob/master/src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html#L11

`<table>` does not have any `slot` attribute, therefore, it will be distributed in the default one, at the very bottom? Is it desired? Do we want app devs to copy'n'paste that?

I think it's better to use explicit slot names as they are well... explicit. If we really need to show-case and test default slot behavior, I'd rather see the separate demo page.

//cc @warpech 

(BTW, it just let be discover a new bug in `sc-include`)

warpech:
It is obviously a bug... or a trap for bugs in `starcounter-include` ;)

Can be fixed next week by Egor.

tomalec:
I have added some unit tests in https://github.com/Juicy/imported-template/pull/44, so hopefully this trap is no longer needed ;)

kegor:
Fixed in #222.

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/KitchenSink/releases/tag/2.0.0
- published in the App Warehouse
- blogged at https://starcounter.io/kitchensink-2-0-0-released-includes-34-common-ui-patterns/

PR #222, opened 2018-01-09
Problem#219

kegor:
Possible fix for #219.


alshakero:
A minor comment, if you add a space before the `#` in the comment message the issue number would be clickable/more accessible. As in `Problem #219: Paginated table does not have explicit slot`.

PR #223, opened 2018-01-09
Fix for firefox failing tests on SC 2.4

PatrykSzwer:
Right now, there are many tests that fail because of `InternalError: Too much recursion`. It's caused by a bug in geckodriver (https://github.com/mozilla/geckodriver/issues/904). This fix should prevent that from happening. If it passes the review, It should be applied in all apps that have selenium tests with `ExecuteScript("return arguments[0].shadowRoot", shadowRootElement);` usage.

### DIFF
--- a/test/KitchenSink.Tests/Ui/BasePage.cs
+++ b/test/KitchenSink.Tests/Ui/BasePage.cs
@@ -114,13 +114,13 @@ namespace KitchenSink.Tests.Ui
             return scriptResult;
         }
 
-        public IWebElement GetElementByQuerySelector(By elementSelector, string queryArgument)
+        public IWebElement GetShadowElementByQuerySelector(By elementSelector, string queryArgument)
         {
             var shadowRootElement = Driver.FindElement(elementSelector);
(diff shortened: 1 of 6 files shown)


test/KitchenSink.Tests/Ui/BasePage.cs
```diff
@@ -106,16 +106,26 @@ public void DblClickOn(IWebElement elementName, int seconds = 10)
...
+            return scriptResult;
+        }
+
+        public IWebElement GetElementByQuerySelector(By elementSelector, string queryArgument)
+        {
+            var shadowRootElement = Driver.FindElement(elementSelector);
+            return GetElementByQuerySelector(shadowRootElement, queryArgument);
+        }
+
+        public IWebElement GetElementByQuerySelector(IWebElement shadowRootElement, string queryArgument)
```
warpech: Can you pls rename this method to `GetShadowElementByQuerySelector` or sth like that. The current name is misleading.

PatrykSzwer: I was actually thinking about something different, but nothing came to my mind. I'll do it right away!

Issue #225, opened 2018-01-09
Upgrade juicy-select to the current version


warpech:
Upgrade `juicy-select` with Bower to the current version https://github.com/Juicy/juicy-select/releases/tag/1.1.0

kegor:
Closed as it is fixed in the #226.

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/KitchenSink/releases/tag/2.0.0
- published in the App Warehouse
- blogged at https://starcounter.io/kitchensink-2-0-0-released-includes-34-common-ui-patterns/

PR #226, opened 2018-01-11
Problem #225: we need to update juicy-select to 1.1.0 version

kegor:
Fix for #225.

There is typo in the issue branch name, right number is 225.

### DIFF
--- a/src/KitchenSink/wwwroot/sys/juicy-select/index.html
+++ b/src/KitchenSink/wwwroot/sys/juicy-select/index.html
@@ -12,6 +12,8 @@
         <template is="dom-bind" id="root">
             <h3 class="lead"><code>juicy-select</code> - Polymer select element with multiple option support</h3>
 
+            <h4>Simple usage</h4>
+
             <div class="horizontal-section">
                 <div>Select countries you have visited:</div>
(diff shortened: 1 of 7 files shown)


src/KitchenSink/wwwroot/sys/juicy-select/index.html
```diff
@@ -22,27 +24,38 @@ <h3 class="lead"><code>juicy-select</code> - Polymer select element with multipl
...
+                    .my-select {
+                        color: blue; 
+                        font-family: Verdana, Tahoma, sans-serif;
+                    }
+                </style>
+            </div>
         </template>
         <script>
-        
+        window.addEventListener('WebComponentsReady', function(e) {
```
warpech: This is not needed in a HTML view imported using `<starcounter-include>`, please remove it.

alshakero: I think this is more of a `juicy-select` issue than a KitchenSink's.

warpech: Please disregard my previous comment here. This is in the demo page.

Issue #227, opened 2018-01-12
Failing tests on Firefox in Starcoutner 2.3


warpech:
There are failing tests on `master` branch after https://github.com/Starcounter/KitchenSink/pull/223 was merged to `master-2.4` and then `master-2.4` to `master`. That was the kind of change that was supposed to be done with [`ours` merge strategy](https://git-scm.com/docs/merge-strategies) on the `master` branch.

@PatrykSzwer can you please revert https://github.com/Starcounter/KitchenSink/pull/223 on the `master` branch? It should only be kept on `master-2.4`

PatrykSzwer:
Closing as done.

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/KitchenSink/releases/tag/2.0.0
- published in the App Warehouse
- blogged at https://starcounter.io/kitchensink-2-0-0-released-includes-34-common-ui-patterns/

PR #228, opened 2018-01-12
Revert changes to firefox tests

PatrykSzwer:
For #227

### DIFF
--- a/src/KitchenSink/wwwroot/sys/pikaday-decorator/.bower.json
+++ b/src/KitchenSink/wwwroot/sys/pikaday-decorator/.bower.json
@@ -37,6 +37,6 @@
     "commit": "f768b99657cf71f938c075c86d930007e9c5c208"
   },
   "_source": "https://github.com/display-interactive/pikaday-decorator.git",
-  "_target": "^0.9.0",
+  "_target": "*",
   "_originalSource": "pikaday-decorator"
 }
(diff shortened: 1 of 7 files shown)


PatrykSzwer:
There is one test fail after this change:
image 
It's related to this: https://github.com/Starcounter/KitchenSink/pull/216#issuecomment-356313474
Now the question is, what should we do about it? @alshakero wasn't sure if he should fix this right now. @warpech WDYT?

warpech:
If I understand correctly, the failing test expects Palindrom 3.1.0, but Starcounter 2.3.1 comes with Palindrom 3.0.9.

In my opinion the failing test should be removed or commented out in KitchenSink, because it won't pass in Starcounter 2.3.1.

PatrykSzwer:
So since 2.3.1 was released we can't do any quick fixes to it? And if I understood @alshakero correctly before, It's not only tests problem, entire functionality stopped working.

warpech:
> So since 2.3.1 was released we can't do any quick fixes to it? And if I understood @alshakero correctly before, It's not only tests problem, entire functionality stopped working.

The functionality did not *stop working*. It never worked in Starcounter 2.3.1.

We can do hotfixes to Starcounter 2.3.1, but the discussed functionality is a new feature. New features should not be added to old releases as hotfixes.

PatrykSzwer:
Since tests passed I'll merge this, to have this change in kitchensink release. If you will have any comments please create new issue for that as a post-review.

Issue #235, opened 2018-01-18
KitchenSink.Tests targetFramework=4.5.2, while for KitchenSink is 4.6.2


kegor:
While exploring KitchenSink.Tests project I found the unexpected detail, that despite the fact that KitchenSink itself has targetFramework=4.6.2 tests still has targetFramework=4.5.2.

I had issue [Use .NET Framework 4.6.1 for apps](https://github.com/Starcounter/Home/issues/347) in a near past, and there is looks reasonable to change targetFramework version to 4.6.2 following the main project?

@warpech are you agree?

warpech:
Yes, please do that. Thanks for suggesting.

kegor:
Closed because fixed in #236.

PR #236, opened 2018-01-19
Problem 235: changing targetFramework to 4.6.1 for both projects in solution

kegor:
Possible fix for #235.

Changes in packages.json were made by manual changing targetFramework version for packages.

### DIFF
--- a/src/KitchenSink/KitchenSink.csproj
+++ b/src/KitchenSink/KitchenSink.csproj
@@ -11,7 +11,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>KitchenSink</RootNamespace>
     <AssemblyName>KitchenSink</AssemblyName>
-    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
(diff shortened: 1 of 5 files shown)


warpech:
Weren't we supposed to use 4.6.1 in the apps? (https://github.com/Starcounter/Home/issues/347)

PatrykSzwer:
TBH I don't see a reason why It shouldn't work with 4.6.2. 
But from the other side, It would be better if We had all apps in same versions. @kegor could You try to make it 4.6.1, not 4.6.2?

kegor:
Hm, I tried to rerun check for 2.4 in TC and on a second time it was successful: https://teamcity.starcounter.org/viewLog.html?buildId=44094&, is this ok case for checking such situations?

warpech:
It is not a matter if it won't work with 4.6.2. The matter is that 4.6.1 comes with VS2017 by default, as far as I understand.

warpech:
> Hm, I tried to rerun check for 2.4 in TC and on a second time it was successful: https://teamcity.starcounter.org/viewLog.html?buildId=44094&, is this ok case for checking such situations?

Yes, as long as the tests are flaky, it is OK to re-run the test until it passes. Let's fix the flaky tests separately from this issue.

warpech:
I can see that I have agreed to that myself in https://github.com/Starcounter/KitchenSink/issues/235#issuecomment-358638362, so I have no further comments :)

PR #241, opened 2018-01-25
Replace bootstrap with underwear.css + uniform.css

PatrykSzwer:
For https://github.com/Starcounter/KitchenSink/issues/238


alshakero:
Am I mistakes or we agreed not to merge for the mean time?

warpech:
We agreed not to merge until Underwear.css is available directly in Starcounter 2.4

This PR is just to increase the visibility and code review friendliness of the branch.

src/KitchenSink/wwwroot/KitchenSink/CheckboxPage.html
```diff
@@ -1,10 +1,10 @@
 <template>
     <dom-bind><template is="dom-bind">
-        <h1 slot="kitchensink/checkbox-heading" class="kitchensink-heading-1">Checkbox</h1>
+        <h1 slot="kitchensink/checkbox-heading">Checkbox</h1>
 
         <p slot="kitchensink/checkbox-description">JSON's primitive value type <code>boolean</code> can be mapped directly to HTML form elements <code>input type="checkbox"</code>.</p>
 
-        <label slot="kitchensink/checkbox-label">Driver's license</label>
+        <span slot="kitchensink/checkbox-label">Driver's license</span>
```
tomalec: Why it's changed from label to span? Isn't it meant to be a label for the following input?

src/KitchenSink/wwwroot/KitchenSink/ClientLocalStatePage.html
```diff
@@ -1,28 +1,29 @@
...
-                        </template>
-                    </hover-observer>
-                </li>
-            </template>
-        </ul>
-    </template></dom-bind>
+    <dom-bind>
+        <template is="dom-bind">
+            <p><em>Please hover an item to see it in action</em></p>
+            <ul class="kitchensink-clientlocalstatepage-hoverable-list">
```
tomalec: Do we still need this class? Can't we do it from Shadow DOM?
We could also add slot attributes to those elements

src/KitchenSink/wwwroot/KitchenSink/DatagridPage.html
```diff
@@ -26,14 +26,16 @@ <h1 slot="kitchensink/datagrid-heading" class="kitchensink-heading-1">Datagrid</
...
+        <button slot="kitchensink/datagrid-button" value="{{model.AddPetTrigger$::click}}" onmousedown="++this.value">Add a pet</button>
     </template></dom-bind>
     <template is="declarative-shadow-dom">
         <slot name="kitchensink/datagrid-heading"></slot>
         <slot name="kitchensink/datagrid-description"></slot>
         <slot name="kitchensink/datagrid-demo-description"></slot>
         <slot name="kitchensink/datagrid-demo-columns"></slot>
         <slot name="kitchensink/datagrid-table"></slot>
-        <slot name="kitchensink/datagrid-button"></slot>
+        <p>
```
tomalec: Why `<p>` to me, it suggests that there will be some text divided into paragraphs. It seems it's just to make `<button>` appear in a new block, isn't it? If so, I'd use just `<div>`

src/KitchenSink/wwwroot/KitchenSink/DialogPage.html
```diff
@@ -1,30 +1,36 @@
...
 
-        <button slot="kitchensink/dialog-button" class="btn btn-sm btn-default" value="{{model.OpenTrigger$::click}}" onmousedown="++this.value">Open dialog box</button>
+        <button slot="kitchensink/dialog-button" value="{{model.OpenTrigger$::click}}" onmousedown="++this.value">Open dialog box</button>
 
         <p slot="kitchensink/dialog-message" class$="{{model.MessageCss}}">{{model.Message}}</p>
 
         <paper-dialog slot="kitchensink/dialog-paper-dialog" id="actions" opened="[[model.Opened]]" modal="true">
+            <style>
+                @import "/sys/uniform.css/uniform.css";
+            </style>
```
tomalec: Are you aware that those styles would affect all apps' light DOM in the solution?
I don't think it's something we should advertise.
It seems like a good use case for `<declarative-shadow-dom>` as an independent element.

src/KitchenSink/wwwroot/KitchenSink/LazyLoadingPage.html
```diff
@@ -30,21 +30,19 @@ <h1 slot="kitchensink/lazyLoading-heading" class="kitchensink-heading-1">Lazy Lo
             </template>
             </tbody>
         </table>
-        <div slot="kitchensink/lazyLoading-message-box">
-            <template is="dom-if" if="{{model.DisplayHoverBox$}}">
-                <div class="kitchensink-lazy-loading__hover-box">
-                    <p>{{model.SelectedPersonsName$}}'s Favorite game</p>
+        <template is="dom-if" if="{{model.DisplayHoverBox$}}">
+            <div slot="kitchensink/lazyLoading-message-box" class="kitchensink-lazy-loading__hover-box">
```
tomalec: I'm not 100% sure, but maybe we could move this `class="kitchensink-lazy-loading__*"` to Shadow DOM as well?

src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
```diff
@@ -3,49 +3,33 @@
...
-                </template>
-                <li class$="{{getLiClassDisabled(model.DisableLast)}}">
-                    <span class="kitchensink-pagination-span" id="kitchensink-pagination-next" tabindex="0" value="{{model.NextPageTrigger$::click}}" on-click="onNextPageClick">Next</span>
-                </li>
-                <li>
-                    <span class="kitchensink-pagination-span" id="kitchensink-pagination-last" tabindex="0" value="{{model.LastPageTrigger$::click}}" on-click="onLastPageClick">Last</span>
-                </li>
-            </ul>
-        </nav>
+        <span slot="kitchensink/pagination-navigation-first" class="previous" id="kitchensink-pagination-first" tabindex="0" value="{{model.FirstPageTrigger$::click}}" on-click="onFirstPageClick" href="">First</span>
```
tomalec: Do we still need this id and class?

src/KitchenSink/wwwroot/KitchenSink/ProgressBarPage.html
```diff
@@ -11,18 +11,19 @@ <h1 slot="kitchensink/progressBar-heading" class="kitchensink-heading-1">Progres
             <span class="kitchensink-progress-bar__percentages"> {{model.Progress}}% </span>
             <div class="kitchensink-progress-bar__fill" style="width:{{model.Progress}}%"></div>
         </div>
-        <p slot="kitchensink/progressBar-button-area">
-            <button class="btn btn-sm btn-default kitchensink-progress-bar__button" value="{{model.StartProgressTrigger$::click}}" onmousedown="++this.value" disabled$="{{model.TaskIsRunnning}}"> {{model.DownloadButtonText}} </button>
-            <template is="dom-if" if="{{model.Progress}}">
-                <span>{{model.FileDownloadText}}</span>
-            </template>
-        </p>
+        <button slot="kitchensink/progressBar-button" class="kitchensink-progress-bar__button" value="{{model.StartProgressTrigger$::click}}" onmousedown="++this.value" disabled$="{{model.TaskIsRunnning}}"> {{model.DownloadButtonText}} </button>
```
tomalec: Can we move `.kitchensink-progress-bar__button` to shadow as well?

src/KitchenSink/wwwroot/KitchenSink/TextareaPage.html
```diff
@@ -1,12 +1,12 @@
 <template>
     <dom-bind><template is="dom-bind">
-        <h1 slot="kitchensink/textarea-heading" class="kitchensink-heading-1">Textarea</h1>
+        <h1 slot="kitchensink/textarea-heading">Textarea</h1>
 
         <p slot="kitchensink/textarea-description">JSON's primitive value type <code>string</code> can be mapped directly to HTML form <code>textarea</code>.</p>
 
-        <label slot="kitchensink/textarea-label" class="control-label kitchensink-test-bio-reaction-label">{{model.BioReaction}}</label>
+        <label slot="kitchensink/textarea-label" class="kitchensink-test-bio-reaction-label">{{model.BioReaction}}</label>
```
tomalec: Can we remove `.kitchensink-test-bio-reaction-label` as well? If it's needed for Selenium, maybe we could use `[slot="kitchensink/textarea-label"]`

src/KitchenSink/wwwroot/KitchenSink/DatagridPage.html
PatrykSzwer: It's more about making a space between table and button then new line. I can replace this with `<br>` for example, but for me <p> was more clean. Do You have any better suggestions?

src/KitchenSink/wwwroot/KitchenSink/DialogPage.html
PatrykSzwer: I'm not sure how I should do that? Can You make an example?

src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
PatrykSzwer: I saw that uniform actually uses those classes, but id can be safely removed.

src/KitchenSink/wwwroot/KitchenSink/TextareaPage.html
PatrykSzwer: All `-test-` classes are created directly for selenium and are used in tests. If You think that It's not good approach You can create separated issue about it. I don't see a point to remove it in one place since everywhere else It's used same way. It will be only more confusing.

src/KitchenSink/wwwroot/KitchenSink/DatagridPage.html
tomalec: `<div style="margin-top: 2em">` seems

src/KitchenSink/wwwroot/KitchenSink/DialogPage.html
tomalec: BTW, do you even use any `uni-class` here?

src/KitchenSink/wwwroot/KitchenSink/ProgressBarPage.html
PatrykSzwer: It's actually not really needed.

src/KitchenSink/wwwroot/KitchenSink/DialogPage.html
tomalec: If you agree that those styles should be scoped, I think the only way to go is the Shadow DOM.
 - I'm not sure if it will work out of the box (may collide with `declarative-shadow-dom` used by `<starcounter-include>`, but does not have to), you can install https://github.com/tomalec/declarative-shadow-dom import it and use as suggested in README.md:
   ```html
   <paper-dialog slot="kitchensink/dialog-paper-dialog" id="actions" opened="[[model.Opened]]" modal="true">
            <template is="declarative-shadow-dom">
                <style>
                    @import "/sys/uniform.css/uniform.css";
                </style>
                <div class="uni-something">
                    <slot name="dialog-header"></slot>
                ...
            </template>
            <h2 slot="dialog-confirmation">Confirmation</h2>
            <button slot="dialog-reject" dialog-dismiss value="{{model.RejectTrigger$::click}}" onmousedown="++this.value;">Reject</button>
            ...
    </paper-dialog>
    ```
- Create above Shadow Root, by yourself:
  - inline script,
  - custom element

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
```diff
@@ -79,13 +105,37 @@ <h2 slot="kitchensink/buttonpage-binding-issue-heading" class="kitchensink-headi
         })();
     </script>
     <template is="declarative-shadow-dom">
+        <style>
+            .kitchensink-add-carrots {
+                display: flex;
+                align-items: baseline;
+            }
+
+            .kitchensink-add-carrots__element { /*TODO: .kitchensink-add-carrots > ::slotted(*) didn't worked'*/
```
tomalec: Is this TODO still applicable?

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
```diff
@@ -20,48 +20,74 @@ <h2 slot="kitchensink/buttonpage-regular-heading" class="kitchensink-heading-2">
 
         <pre slot="kitchensink/buttonpage-regular-reaction-label" class="kitchensink-test-carrots-reaction-label">{{model.CarrotsReaction}}</pre>
 
-        <p slot="kitchensink/buttonpage-regular-add-carrots" class="kitchensink-add-carrots">
-            <span class="kitchensink-add-carrots__label">Add carrots:</span>
+        <span slot="kitchensink/buttonpage-regular-add-carrots-label" class="kitchensink-add-carrots__label">Add carrots:</span>
```
tomalec: Could you move `.kitchensink-add-carrots__label` to the shadow DOM?

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
```diff
@@ -20,48 +20,74 @@ <h2 slot="kitchensink/buttonpage-regular-heading" class="kitchensink-heading-2">
...
+        <span slot="kitchensink/buttonpage-regular-add-carrots-span-function" on-click="onAddCarrotsClick">Span (function)</span>
 
-        <h2 slot="kitchensink/buttonpage-switch-heading" class="kitchensink-heading-2">Switch button</h2>
+        <h2 slot="kitchensink/buttonpage-switch-heading">Switch button</h2>
 
         <p slot="kitchensink/buttonpage-switch-description">When you have a special need to use the pressed state of a button, you might want to use <code>boolean</code> type for buttons.
         </p>
 
-        <button slot="kitchensink/buttonpage-switch-button" class="btn btn-sm btn-default kitchensink-toggle-button" pressed$="{{model.EnableCarrotEngine$}}" on-click="onEnableCarrotEngineClick">Enable carrot engine</button>
+        <button slot="kitchensink/buttonpage-switch-button" class="kitchensink-toggle-button" pressed$="{{model.EnableCarrotEngine$}}" on-click="onEnableCarrotEngineClick">Enable carrot engine</button>
```
tomalec: Could you also move `.kitchensink-toggle-button` styles to shadow dom?

src/KitchenSink/wwwroot/KitchenSink/LazyLoadingPage.html
```diff
@@ -30,21 +30,15 @@ <h1 slot="kitchensink/lazyLoading-heading" class="kitchensink-heading-1">Lazy Lo
...
-                    </template>
-
-                    <template is="dom-if" if="{{!showLoadSpinner(model.DisplayedData.DataContent$)}}">
-                        <p>{{model.DisplayedData.DataContent$}}</p>
-                    </template>
-                </div>
+        <template is="dom-if" if="{{model.DisplayHoverBox$}}">
+            <p slot="kitchensink/lazyLoading-message-box-game-label">{{model.SelectedPersonsName$}}'s Favorite game</p>
+            <template is="dom-if" if="{{showLoadSpinner(model.DisplayedData.DataContent$)}}">
+                <div slot="kitchensink/lazyLoading-message-box-game-loading"></div>
```
tomalec: It seems strange to me that by default, to content here is an empty, meaningless `div`. I think it should have at least the title, and content like "Loading", which then could be used/hidden in Shadow DOM. Plus maybe `aria-busy="true"`

src/KitchenSink/wwwroot/KitchenSink/LazyLoadingPage.html
```diff
@@ -95,10 +89,42 @@ <h1 slot="kitchensink/lazyLoading-heading" class="kitchensink-heading-1">Lazy Lo
         })();
     </script>
     <template is="declarative-shadow-dom">
+        <style>
+            .kitchensink-lazy-loading__spinning-loading::slotted(div) {
```
tomalec: As we are just rendering spinner, regardless of the content, I think `::slotted(*)` would be better.

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
PatrykSzwer: Your example is different than actual use-case. It will work when there are only 2 levels

tomalec:
I'll take over the remaining two bugs:
- [ ] lack of content for loading preloader https://github.com/Starcounter/KitchenSink/pull/241#discussion_r164523601
- [ ] class used in light dom https://github.com/Starcounter/KitchenSink/pull/241#discussion_r164521533,
- [ ] resolve conflicts with master

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
warpech: In this PR, it is enough to just remove Bootstrap.

Long term I would like to go to the extreme and show all that without any custom styles in light DOM. If we show that this is possible, devs might want to follow.

tomalec:
@warpech, @PatrykSzwer  could you review?
It should work fine in polyfilled browsers if you use unminified uniform.css https://github.com/Starcounter/uniform.css/issues/9#issuecomment-364141306

tomalec:
AFAIK, this should be considered as success https://teamcity.starcounter.org/viewLog.html?buildId=46368&tab=buildResultsDiv&buildTypeId=Apps_KitchenSink_KitchenSinkTestFeatureBranch

warpech:
I will review tomorrow. The build `KitchenSink Master branch with Starcounter RC (2.3.1) (KitchenSink)` [fails](https://teamcity.starcounter.org/viewLog.html?buildId=46443&buildTypeId=Apps_KitchenSink_Develop), because of a Git conflict in the datepicker page. 

`vaadin-date-picker` is not present in `master` branch, as it doesn't work with Polymer 1.

ANYWAY, we will be not merging this branch to `master`, because Underwear and Uniform is not supported in Starcounter 2.3.

----

TL;DR only look at the build `KitchenSink Master-2.4 branch with Starcounter Develop (2.4)`

I will review tomorrow.

src/KitchenSink/wwwroot/KitchenSink/BreadcrumbPage.html
```diff
@@ -44,24 +43,29 @@ <h5>Adding a new type in <strong>{{model.CurrentTreeItem.ParentName}}</strong></
...
+        </div>
         <slot name="kitchensink/breadcrumb-navigation-breadcrumb-heading"></slot>
         <slot name="kitchensink/breadcrumb-navigation-breadcrumb-description"></slot>
         <slot name="kitchensink/breadcrumb-navigation-breadcrumb"></slot>
         <slot name="kitchensink/breadcrumb-adding-heading"></slot>
         <slot name="kitchensink/breadcrumb-editing-heading"></slot>
         <slot name="kitchensink/breadcrumb-name-heading"></slot>
         <slot name="kitchensink/breadcrumb-name-input"></slot>
-        <slot name="kitchensink/breadcrumb-name-save"></slot>
+        <slot class="uni-button-success" name="kitchensink/breadcrumb-name-save"></slot>
```
warpech: This should be used in a wrapper.

src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
```diff
@@ -78,21 +62,26 @@ <h1 slot="kitchensink/pagination-heading" class="kitchensink-heading-1">Paginati
...
     <template is="declarative-shadow-dom">
+        <style>
+            @import "/sys/uniform.css/uniform.css";
+        </style>
         <slot name="kitchensink/pagination-heading"></slot>
         <slot name="kitchensink/pagination-description"></slot>
         <slot name="kitchensink/pagination-juicy-select-items-per-page"></slot>
         <slot name="kitchensink/pagination-table"></slot>
-        <slot name="kitchensink/pagination-navigation"></slot>
+        <div class="uni-pagination">
```
warpech: This is not implemented according to Uniform.css 0.0.3. It does not render correctly.

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
```diff
@@ -79,13 +105,41 @@ <h2 slot="kitchensink/buttonpage-binding-issue-heading" class="kitchensink-headi
...
+                display: flex;
+                align-items: baseline;
+            }
+
+            .kitchensink-add-carrots ::slotted(*) {
+                margin-right: 10px !important;
+            }
+            /* Duplicate rule above with selector to support polyfilled browsers and workaround ShadyCSS bug */
+            .kitchensink-add-carrots > * {
+                margin-right: 10px;
```
warpech: If this rule is duplicated, it should include `!important`.

src/KitchenSink/wwwroot/KitchenSink/FileUploadPage.html
```diff
@@ -29,50 +28,47 @@ <h2 slot="kitchensink/fileupload-upload-file-header" class="kitchensink-heading-
...
                             <td>
-                                <button type="button" class="btn btn-sm btn-default" value="{{item.DeleteTrigger$::click}}" onmouseup="++this.value;">Delete</button>
+                                <button type="button" value="{{item.DeleteTrigger$::click}}" onmouseup="++this.value;">Delete</button>
                             </td>
                         </tr>
                     </template>
                     </tbody>
                 </table>
-                <div class="alert alert-warning">
+                <div slot="kitchensink/fileupload-files-table-message" class="kitchensink-test-warning alert alert-warning">
```
warpech: `kitchensink-test-warning` should be removed and changed to `[slot=...]` in the test suite.

Classes `alert` and `alert-xxx` are Bootstrap and should be changed to `uni-alert` in this and other files.

src/KitchenSink/wwwroot/KitchenSink/FileUploadPage.html
```diff
@@ -81,9 +77,9 @@ <h2 slot="kitchensink/fileupload-uploading-header" class="kitchensink-heading-2"
 
         <starcounter-upload slot="kitchensink/fileupload-upload" session-id="{{model.SessionId}}" upload-url="{{model.UploadUrl}}" tasks="{{tasks}}" multiple auto-upload></starcounter-upload>
 
-        <button slot="kitchensink/fileupload-select-button" type="button" class="btn btn-sm btn-primary" on-click="selectFile">Select file</button>
+        <button slot="kitchensink/fileupload-select-button" type="button" on-click="selectFile">Select file</button>
```
warpech: In many files, I can see `<button type="button">`. I think that the `type` attribute can be removed in this case?

src/KitchenSink/wwwroot/KitchenSink/LazyLoadingPage.html
```diff
@@ -95,10 +84,58 @@ <h1 slot="kitchensink/lazyLoading-heading" class="kitchensink-heading-1">Lazy Lo
         })();
     </script>
     <template is="declarative-shadow-dom">
+        <style>
+            .kitchensink-lazy-loading__hover-box {
```
warpech: Actually, I was not expecting this in Shadow DOM. I mean, we can have it for fun, but this is the most extreme case. As I understand, if we can have this in Shadow DOM, we can have everything from `style.css`.

Good job!

src/KitchenSink/wwwroot/KitchenSink/DatepickerPage.html
```diff
@@ -13,29 +13,35 @@ <h1 slot="kitchensink/datepicker-heading" class="kitchensink-heading-1">Datepick
...
+            <slot name="kitchensink/datepicker-month-label"></slot>
+            <slot name="kitchensink/datepicker-month-input"></slot>
+        </p>
+        <p>
+            <slot name="kitchensink/datepicker-day-label"></slot>
+            <slot name="kitchensink/datepicker-day-input"></slot>
+        </p>
     </template>
-</template>
+</template>
```
warpech: The datepicker page does not render correctly for me using Uniform.css 0.0.3:

image

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
tomalec: I don't get it, why?

src/KitchenSink/wwwroot/KitchenSink/LazyLoadingPage.html
tomalec: I believe we can have everything from `style.css` but we decided not to spend to much time on this PR any longer, and close removing bootstrap first.

src/KitchenSink/wwwroot/KitchenSink/FileUploadPage.html
tomalec: We had `alert` classes in uniform, https://github.com/Starcounter/uniform.css/blob/master/src/uniform.css#L246

That's why we need to merge it fast, as it starts to rotten ;)

src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
tomalec: I cannot reproduce the result I see in your screenshot, however in my case it was also not the structure as in demo. BTW, pagination styles are still buggy, but that should be fixed in uniform.css itself.

src/KitchenSink/wwwroot/KitchenSink/FileUploadPage.html
tomalec: TIL: it's better to keep this `<button type="button">` , in case someone will wrap a partial view insertion point in a `<form>` https://stackoverflow.com/questions/41904199/whats-the-point-of-button-type-button

> The first one [`<button>`] is assumed to be `type=submit` since a `type` attribute hasn't been specified.
>
> If you are not in a `<form>` element, the button won't have anything to submit, so it doesn't matter as much. :)
>
> Semantics usually become important at some point in your application's lifetime, though, so it's a good idea to make a habit of specifying the `type`.

src/KitchenSink/wwwroot/KitchenSink/DatepickerPage.html
tomalec: I cannot reproduce it, for me it looks like that even in Firefox 
image

src/KitchenSink/wwwroot/KitchenSink/FileUploadPage.html
alshakero: I spent years using `<input type="button">` to avoid auto-submit behaviour. Never knew about `<button type="button">` solution!

src/KitchenSink/wwwroot/KitchenSink/LazyLoadingPage.html
warpech: We should not merge this PR before SCF 3.x is included in Starcounter. I agree that this PR can be soon marked as "ready to merge". We will then continue working by targetting new PRs to this PR.

src/KitchenSink/wwwroot/KitchenSink/FileUploadPage.html
warpech: In Uniform we have `.uni-alert`, in KitchenSink I can still find `.alert` in light DOM.

src/KitchenSink/wwwroot/KitchenSink/ButtonPage.html
warpech: Because "duplication" means we have exactly the same rule. Now we have a slightly different rule for polyfilled browsers.

tomalec: Ah now I get what you mean, but still... We would like to have "the rule that have same effect as >exactly< as we can". Maybe we should change the "Deuplicate" word. Making shadow root rule `!important` make it more important than  regular light DOM rule, but less important than any other light DOM `!important` rule.

If we would add `!important` as you suggested for polyfilled browsers we may jump higher in priority list.

src/KitchenSink/wwwroot/KitchenSink/LazyLoadingPage.html
tomalec: I'm not so sure if it's the easiest way to maintain it. This PR is already old and long and now becomes a PR to fix all the things in Kitchensink.

Maybe we can merge the PR, but to keep the branches separate. So we will not have to review all those things again when we will have to eventually close this PR + 10 new ones merge into it.

src/KitchenSink/wwwroot/KitchenSink/FileUploadPage.html
tomalec: Wow, not only in light DOM in some HTML files, it's used in data fed to HTML via data binding.

tomalec: For `.kitchensink-test-*` I'd reopen https://github.com/Starcounter/KitchenSink/issues/242 and create a separate PR as it has nothing to do with bootstrap nor uniform

src/KitchenSink/wwwroot/KitchenSink/DatepickerPage.html
PatrykSzwer: I confirm what @tomalec, no reproduction for me either:

warpech: My bad... Now it works ok for me, too

alshakero:
One minor but pet-peevy issue. `CheckboxPage` has this:

```html
<input slot="kitchensink/checkbox-input" type="checkbox" checked="{{model.DrivingLicense$::change}}">

<div slot="kitchensink/checkbox-license-label">{{model.DrivingLicenseReaction}}</div>
```

Which results in this:
image

This HTML would fix it:
```html
<div slot="kitchensink/checkbox-license-label"><input slot="kitchensink/checkbox-input" type="checkbox" checked="{{model.DrivingLicense$::change}}"> {{model.DrivingLicenseReaction}}</div>
```

Can I push this commit?

tomalec:
> Which results in this: ....

Do you mean this is not the desired look?  What is the desired, then?

alshakero:
Not for Checkbox 
image

PR #252, opened 2018-02-19
juicy-html Bug Fix

atifwaqar:


alshakero:
Thank you so much Atif! Could you change the destination branch to `master-2.4`. `master`  still has Polymer 1.x code where `template is=` works.

PR #254, opened 2018-02-20
Make use of Uniform.css styling on the form control pages

warpech:
@PatrykSzwer can you perhaps take over from here and go through it with @andwah? KitchenSink should show the usage of Uniform.css, as a source for copy-paste driven development.


warpech:
Ooops, I based it on a wrong branch (`master`)

Issue #260, opened 2018-03-08
Use new classes from uniform in Kitchensink


PatrykSzwer:

warpech:
I will push a few improvements today

PatrykSzwer:
Closing as done.

PR #261, opened 2018-03-08
Advanced uniform classes

PatrykSzwer:
For #260


warpech:
I pushed some fixes. @PatrykSzwer pls merge if you agree.

I want to fix the formatting of the side navigation on the left, but I will post it as a separate PR.

Issue #269, opened 2018-03-26
DatepickerPage_SelectDate test sometimes fails


warpech:
Sometimes the build fails on the `DatepickerPage_SelectDate` test. For example: https://teamcity.starcounter.org/viewLog.html?buildId=50715&buildTypeId=Apps_KitchenSink_KitchenSinkMaster24branchWithStarcounterDevelopNightly

@kegor could you pls investigate that?

<img width="1139" alt="screen shot 2018-03-26 at 08 58 47" src="https://user-images.githubusercontent.com/566463/37891327-359c22f4-30d4-11e8-9e54-47cfc638b2d0.png">

kegor:
I think that the problem here is in time that datepicker spends on changing the default date on the selected one in the test:
![datepicker_selectdate_test](https://user-images.githubusercontent.com/17431807/37901249-5b8b87f0-30f9-11e8-8c84-5beb06c51f2a.gif)
I think we can either replace `Assert`'s on `WaitUntil` or add one `WaitUntil` before all `Assert`'s, it will wait while datepicker will be closed again.

kegor:
Closed because initial fixes were merged in #270, and new #273 was opened.

PR #270, opened 2018-03-26
Problem #269: fix for flaky DatepickerPage_SelectDate test.

kegor:
**Problem #269:**
image

**Solution:**
To avoid flaky behaviour `Asserts`'s were replaced on `WaitUntil`'s statements.

### DIFF
--- a/test/KitchenSink.Tests/Test/DatepickerPageTest.cs
+++ b/test/KitchenSink.Tests/Test/DatepickerPageTest.cs
@@ -33,9 +33,9 @@ namespace KitchenSink.Tests.Test
             WaitUntil(x => _datePicker.DatePicker.Displayed);
 
             _datePicker.SelectDate("01/01/2016");
-            Assert.AreEqual("2016", _datePicker.YearInput.GetAttribute("value"));
-            Assert.AreEqual("January", _datePicker.MonthInput.GetAttribute("value"));
-            Assert.AreEqual("1", _datePicker.DayInput.GetAttribute("value"));
+            WaitUntil(x => _datePicker.YearInput.GetAttribute("value") == "2016");
+            WaitUntil(x => _datePicker.MonthInput.GetAttribute("value") == "January");
+            WaitUntil(x => _datePicker.DayInput.GetAttribute("value") == "1");
         }
     }
 }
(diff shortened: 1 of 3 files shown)


kegor:
Actually, I see two approaches here:
1. Fixing next flaky tests when they will appear in TC.
2. Replacing all `Asserts` on `WaitUntil` in appropriate cases.
@warpech, @un-tone wdyt?

un-tone:
If we can't confidently rely on `Assert`s in Selenium tests we should use `WaitUntil` in all such places where we do checking UI elements after some action.

un-tone:
Otherwise, there always will appear flaky failed tests, here and there.

warpech:
My problem with `WaitUntil` is that, as opposed to `Assert`, it gives very unclear error messages, such as:

```
OpenQA.Selenium.WebDriverTimeoutException : Timed out after 10 seconds
   at OpenQA.Selenium.Support.UI.DefaultWait`1.ThrowTimeoutException(String exceptionMessage, Exception lastException)
   at OpenQA.Selenium.Support.UI.DefaultWait`1.Until[TResult](Func`2 condition)
   at BlendingEditor.Tests.Ui.BasePage.WaitForElementToBeClickable(IWebElement elementName, Int32 seconds) in C:\TeamCity\BuildAgent\work\1ee82139cef36e7f\Blending\test\BlendingEditor.Tests\Ui\BasePage.cs:line 21
   at BlendingEditor.Tests.Ui.BasePage.ClickOn(IWebElement elementName, Int32 seconds) in C:\TeamCity\BuildAgent\work\1ee82139cef36e7f\Blending\test\BlendingEditor.Tests\Ui\BasePage.cs:line 26
   at BlendingEditor.Tests.Ui.LayoutGroupsPage.ClickButtonCreateLayout(IWebElement button) in C:\TeamCity\BuildAgent\work\1ee82139cef36e7f\Blending\test\BlendingEditor.Tests\Ui\LayoutGroupsPage.cs:line 125
   at BlendingEditor.Tests.Test.LayoutGroupsPageTest.LayoutGroupsPage_AddLayout() in C:\TeamCity\BuildAgent\work\1ee82139cef36e7f\Blending\test\BlendingEditor.Tests\Test\LayoutGroupsPageTest.cs:line 76
```

Any idea how to improve that? [This thread](https://stackoverflow.com/questions/10814807/how-to-improve-error-message-for-webdriver-expectedcondition) suggests two solutions:

1. generic helper method that uses `try`/`catch`
2. `withMessage` method from Selenium API

kegor:
@warpech, If we decide to use common checks in tests (like title text check), then the first option looks more suitable, if we choose the second option, we will have to describe the same text in each test that checks the title, right?

@un-tone wdyt?

kegor:
Hi @warpech, what do you think, do you agree with @un-tone?

PR #277, opened 2018-04-16
Fix pagination (next+prev buttons)

### LINKED ISSUE #248: Clicking next page in pagination advances two pages not one
##### Browsers affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari

alshakero:
Fix #248

### DIFF
--- a/src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/PaginationPage.html
@@ -24,45 +24,19 @@
             </template>
             </tbody>
         </table>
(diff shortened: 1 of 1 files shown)


tomalec:
Have you tried otherwise, to remove https://github.com/Starcounter/KitchenSink/pull/277/files#diff-515dd804e01781397feba2f69f2b2d37R41 and leave just polymer biding? You know less code => the better. As we aim for super-easy

alshakero:
With all honesty I'm surprised it works now. But it does and it's clean.

alshakero:
This is a side effect of using `spans` instead of `buttons`. See https://github.com/Starcounter/KitchenSink/issues/263.

PatrykSzwer:
Ok, then I think It's ready to be merged.

PR #285, opened 2018-04-22
Add deployment instructions

warpech:
for kitchensink.starcounter.io 

for https://github.com/Starcounter/KitchenSink/issues/205#issuecomment-383409767

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,17 @@ To contribute code to this repository, follow the instructions in the [guideline
 
 To release the app to the warehouse, follow the instructions in the [guidelines](https://github.com/Starcounter/CompanyTrack/blob/master/AppsTeam/Guidelines/releasing-to-warehouse.md).
 
+## How to deploy to `kitchensink.starcounter.io`
+
+1. Make sure that the new version of KitchenSink is released to the App Warehouse
+2. Connect via RDP to 54.194.101.35 (credentials in `N*********.kdbx`)
+3. In RDP, wipe out the current data by running `staradmin delete db default` 
+4. Install the current version of Starcounter 2.4 from http://downloads.starcounter.com/
+5. Start Starcounter by running `staradmin start server`
+6. Create a new database on port 80 by running `staradmin new db default DefaultUserHttpPort=80`
+7. In Administrator, go to the database `default` and install the current version of KitchenSink from App Warehouse
+8. Go to https://kitchensink.starcounter.io/ to verify that the new KitchenSink is correctly deployed
+
 ## Testing
 
 ### Prepare your environment


CONTRIBUTING.md
```diff
@@ -40,6 +40,17 @@ To contribute code to this repository, follow the instructions in the [guideline
 
 To release the app to the warehouse, follow the instructions in the [guidelines](https://github.com/Starcounter/CompanyTrack/blob/master/AppsTeam/Guidelines/releasing-to-warehouse.md).
 
+## How to deploy to `kitchensink.starcounter.io`
+
+1. Make sure that the new version of KitchenSink is released to the App Warehouse
+2. Connect via RDP to 54.194.101.35 (credentials in `N*********.kdbx`)
+3. In RDP, wipe out the current data by running `staradmin delete db default`
```
PatrykSzwer: Is this step really necessary? What could happen in DB that required removing it every time we update version of KS?

CONTRIBUTING.md
```diff
@@ -40,6 +40,17 @@ To contribute code to this repository, follow the instructions in the [guideline
...
 To release the app to the warehouse, follow the instructions in the [guidelines](https://github.com/Starcounter/CompanyTrack/blob/master/AppsTeam/Guidelines/releasing-to-warehouse.md).
 
+## How to deploy to `kitchensink.starcounter.io`
+
+1. Make sure that the new version of KitchenSink is released to the App Warehouse
+2. Connect via RDP to 54.194.101.35 (credentials in `N*********.kdbx`)
+3. In RDP, wipe out the current data by running `staradmin delete db default` 
+4. Install the current version of Starcounter 2.4 from http://downloads.starcounter.com/
+5. Start Starcounter by running `staradmin start server`
+6. Create a new database on port 80 by running `staradmin new db default DefaultUserHttpPort=80`
```
PatrykSzwer: I think this step is not really necessary since running kitchensink for the first time without `default` database, will create it automatically.

CONTRIBUTING.md
warpech: It is not necessary, but can help to avoid problems. 

In case of KitchenSink, we don't care about the stored data, so it is easier to delete the database than to fix possible problems with data migration to a newer version.

Typically in hosted demos like KitchenSink, the database is purged periodically, e.g. once a day or once an hour.

Do you think I should remove this step?

PatrykSzwer: > Typically in hosted demos like KitchenSink, the database is purged periodically, e.g. once a day or once an hour.

CONTRIBUTING.md
```diff
@@ -40,6 +40,17 @@ To contribute code to this repository, follow the instructions in the [guideline
 
 To release the app to the warehouse, follow the instructions in the [guidelines](https://github.com/Starcounter/CompanyTrack/blob/master/AppsTeam/Guidelines/releasing-to-warehouse.md).
 
+## How to deploy to `kitchensink.starcounter.io`
+
+1. Make sure that the new version of KitchenSink is released to the App Warehouse
```
tomalec: Can we have a link to the release to Warehouse process here?

alshakero: You can use this @PatrykSzwer https://github.com/Starcounter/CompanyTrack/blob/master/AppsTeam/Guidelines/releasing-to-warehouse.md

warpech: The link is on line 41

PR #286, opened 2018-04-22
Refresh screenshot

warpech:
Update the screenshot to the current version
Move it to the top of the file, make it clickable as the link to the demo
Remove the outdated link to a YouTube video

For https://github.com/Starcounter/KitchenSink/issues/205#issuecomment-383409767


warpech:
I think that's because links with `target="blank"` are considered a usability and accessibility anti-pattern. See https://stackoverflow.com/questions/3906898/to-target-blank-or-not-to-target-blank-that-is-the-question

Issue #164, opened 2017-08-10
Add inline code examples to pages


rogeralsing:
I would like to see inline code snippets for each example.
Similar to how it is displayed here: https://www.webcomponents.org/element/PolymerElements/paper-input/demo/demo/index.html

And paths to what source files in the solution/project are used per example.
e.g.

Page:

EXAMPLE

Snippet with the relevant stuff

Links to the full code files

alshakero:
I think we can automate this, a small expandable custom element that shows the same code of its `host` in a syntax highlighter. And throw this CE in every view.

rogeralsing:
Also, maybe consider making the highlighter an actual code editor, allow the user to tinker around with the client-side code and see what's going on. might be a much bigger task but it would be pretty neat.
(webcomponents.org site does this too)

alshakero:
@warpech could you make one single example of this? I tried with it a little and I don't really understand what we want out of it. If you could make a single page example I'd really appreciate it.

warpech:
The absolute minimum for now is to add links to source code for HTML, JSON and JSON.CS files that make the triplet. I imagine it like this:

Add a `<kitchensink-page-links page="TextPage">` to each view.

`kitchensink-page-links` should render three GitHub links, for HTML, JSON and JSON.CS representations of the `page`.

You can use the version number from https://github.com/Starcounter/KitchenSink/issues/284 to create the GitHub link to the source code relevant for the given version of the app. Before #284 is done, you can just hardcode `master-2.4`.

alshakero:
> Add a <kitchensink-source-links></kitchensink-source-links> at the bottom of each view.

I don't think this is needed. Here is what I did in `MasterPage.html`.

```html
<dom-bind>
    <template id="root" is="dom-bind">
        <imported-template href="{{model.NavPage.KitchenSink_0.Html}}"></imported-template>
        <div slot="kitchensink/current">
            <starcounter-include view-model="{{model.CurrentPage}}"></starcounter-include>
            <div id="code-sample-wrapper">
                <kitchensink-source-links url$="{{model.CurrentPage.KitchenSink_0.Html}}"></kitchensink-source-links>
            </div>
        </div>
    </template>
</dom-bind>
```

warpech:
> I don't think this is needed. Here is what I did in `MasterPage.html`.

[...]

PR #289, opened 2018-04-25
WIP: Add a sketch of a way to automatically include code snippets

tomalec:
https://github.com/Starcounter/KitchenSink/issues/164

Use custom element, that will fetch the HTML source from imported document,
Later to be able to fetch it from the sourcefiles


warpech:
Yesterday we agreed on the meeting that this won't be needed because for #164 it is enough to add links and #293 will require source code of all three files (HTML, JSON, JSON.CS)

alshakero:
This is redundant to #301 right?

tomalec:
Right :)

PR #294, opened 2018-04-26
Fix styles warnings #282

### LINKED ISSUE #282: Fix warning about styling master document from HTML Import
In Chrome, currently I get the following error:

image

Solve this according to information in https://github.com/Starcounter/RebelsLounge/issues/222#issuecomment-382777626

alshakero:
Some warnings are still showing due to custom elements doing the same mistake, such as Vaadin date picker. You can visit https://vaadin.com/components/vaadin-date-picker and you'll see the same warning showing

Fix #282


tomalec:
hm.. interesting, personally I would say preloading styles as any other content is a legit thing to do even though attaching this styles in not nice.

`deduped-link` https://github.com/Starcounter/RebelsLounge/issues/234 could solve that, but for sure this way proposed here will work sooner and simpler.

My doubts regarding this solution:
- doesn't it introduce/increase the FOUC, as this is no fetched later,
- does it make any sense to preload something that should be already there when you are stamping content,
- uniform.css should not be a part of a content, but composition.

Anyway I'd get a deeper look into the spec to check if and why preload is also wrong.

Issue #284, opened 2018-04-22
Add information about the current version of the app


warpech:
On the main page, add information about the current version of the app (ideally take it from assembly info). 

This information will allow to easily see if the version deployed at https://kitchensink.starcounter.io/ is the current version of KitchenSink.

alshakero:
Done in https://github.com/Starcounter/KitchenSink/pull/301

PR #301, opened 2018-05-03
Show app and Starcouner version, and add source code links

### LINKED ISSUE #164: Add inline code examples to pages

alshakero:
Fix #164 and #284 

For you own comfort, please review per commit, I isolated the commits carefully.

### DIFF
--- a/src/KitchenSink/MainPage.json
+++ b/src/KitchenSink/MainPage.json
@@ -1,3 +1,5 @@
 {
-  "Html": "/KitchenSink/MainPage.html"
+  "Html": "/KitchenSink/MainPage.html",
+  "appVersion": "",
+  "starcounterVersion":  ""
 }
--- a/src/KitchenSink/MasterPage.json
+++ b/src/KitchenSink/MasterPage.json
@@ -1,5 +1,7 @@
 {
   "Html": "/KitchenSink/MasterPage.html",
   "NavPage": {},
-  "CurrentPage": {}
+  "CurrentPage": {},
+  "appVersion": "",
+  "starcounterVersion": ""
 }
--- a/src/KitchenSink/Program.cs
+++ b/src/KitchenSink/Program.cs
@@ -7,6 +7,16 @@ namespace KitchenSink
 {
     class Program
     {
+        static string getAppVersionFromAssemblyFile()
+        {
+            System.Reflection.Assembly assembly = System.Reflection.Assembly.GetExecutingAssembly();
+            System.Diagnostics.FileVersionInfo fvi = System.Diagnostics.FileVersionInfo.GetVersionInfo(assembly.Location);
+           return fvi.FileVersion;
+        }
+        static string getStarcounterVersion()
+        {
+            return Starcounter.Internal.CurrentVersion.Version;
+        }
         static void Main()
         {
             var app = Application.Current;
@@ -17,7 +27,14 @@ namespace KitchenSink
 
             Handle.GET("/KitchenSink/json", () => new Json());
 
-            Handle.GET("/KitchenSink/partial/mainpage", () => new MainPage());
+            Handle.GET("/KitchenSink/partial/mainpage", () => {
+                return new MainPage
+                {
+                    starcounterVersion = getStarcounterVersion(),
+                    appVersion = getAppVersionFromAssemblyFile()
+                };
+            });
+
             Handle.GET("/KitchenSink/mainpage", () => WrapPage<MainPage>("/KitchenSink/partial/mainpage"));
 
             Handle.GET("/KitchenSink", () => Self.GET("/KitchenSink/mainpage"));
@@ -276,6 +293,8 @@ namespace KitchenSink
             {
                 master = new MasterPage
                 {
+                    appVersion = getAppVersionFromAssemblyFile(),
+                    starcounterVersion = getStarcounterVersion(),
                     NavPage = Self.GET("/kitchensink/nav")
                 };
                 Session.Current.Store[nameof(MasterPage)] = master;
--- a/src/KitchenSink/wwwroot/KitchenSink/MainPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/MainPage.html
@@ -7,7 +7,9 @@
             <h1 slot="kitchensink/main-heading">Welcome to KitchenSink</h1>
             <p slot="kitchensink/main-description">In this app you can take a look on typical UI patterns in a form that is easy to apply in any Starcounter app.</p>
             <p slot="kitchensink/main-description">Use the menu on the left to browse various examples of how to build web-based UI with Starcounter and Polymer.</p>
-            <p slot="kitchensink/main-app-version"><em>App version: {{model.appVersion}}, running on Starcounter: {{model.starcounterVersion}}</em></p>
+            <p slot="kitchensink/main-app-version">
+                <em>App version: {{model.appVersion}}, running on Starcounter: {{model.starcounterVersion}}</em>
+            </p>
             <github-source-links></github-source-links>
         </template>
     </dom-bind>
--- a/src/KitchenSink/wwwroot/KitchenSink/MainPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/MainPage.html
@@ -1,5 +1,5 @@
 <link rel="preload" as="style" href="/sys/uniform.css/uniform.css">
-
+<link rel="import" href="/sys/github-source-links/github-source-links.html">
 
 <template>
     <dom-bind>
@@ -8,6 +8,7 @@
             <p slot="kitchensink/main-description">In this app you can take a look on typical UI patterns in a form that is easy to apply in any Starcounter app.</p>
             <p slot="kitchensink/main-description">Use the menu on the left to browse various examples of how to build web-based UI with Starcounter and Polymer.</p>
             <p slot="kitchensink/main-app-version"><em>App version: {{model.appVersion}}, running on Starcounter: {{model.starcounterVersion}}</em></p>
+            <github-source-links></github-source-links>
         </template>
     </dom-bind>
     <template is="declarative-shadow-dom">
--- a/src/KitchenSink/wwwroot/KitchenSink/MainPage.html
+++ b/src/KitchenSink/wwwroot/KitchenSink/MainPage.html
@@ -1,17 +1,22 @@
 <link rel="preload" as="style" href="/sys/uniform.css/uniform.css">
 
-<template>
-    <h1 slot="kitchensink/main-heading">Welcome to KitchenSink</h1>
-    
-    <p slot="kitchensink/main-description">In this app you can take a look on typical UI patterns in a form that is easy to apply in any Starcounter app.</p>
-    <p slot="kitchensink/main-description">Use the menu on the left to browse various examples of how to build web-based UI with Starcounter and Polymer.</p>
 
+<template>
+    <dom-bind>
+        <template is="dom-bind">
+            <h1 slot="kitchensink/main-heading">Welcome to KitchenSink</h1>
+            <p slot="kitchensink/main-description">In this app you can take a look on typical UI patterns in a form that is easy to apply in any Starcounter app.</p>
+            <p slot="kitchensink/main-description">Use the menu on the left to browse various examples of how to build web-based UI with Starcounter and Polymer.</p>
+            <p slot="kitchensink/main-app-version"><em>App version: {{model.appVersion}}, running on Starcounter: {{model.starcounterVersion}}</em></p>
+        </template>
+    </dom-bind>
     <template is="declarative-shadow-dom">
         <link rel="stylesheet" href="/sys/uniform.css/uniform.css">
         <div class="uni-section-primary">
             <div class="uni-card">
                 <slot name="kitchensink/main-heading"></slot>
                 <slot name="kitchensink/main-description"></slot>
+                <slot name="kitchensink/main-app-version"></slot>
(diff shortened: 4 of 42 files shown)


src/KitchenSink/wwwroot/KitchenSink/MainPage.html
```diff
@@ -1,17 +1,22 @@
...
-    
-    <p slot="kitchensink/main-description">In this app you can take a look on typical UI patterns in a form that is easy to apply in any Starcounter app.</p>
-    <p slot="kitchensink/main-description">Use the menu on the left to browse various examples of how to build web-based UI with Starcounter and Polymer.</p>
 
+<template>
+    <dom-bind>
+        <template is="dom-bind">
+            <h1 slot="kitchensink/main-heading">Welcome to KitchenSink</h1>
+            <p slot="kitchensink/main-description">In this app you can take a look on typical UI patterns in a form that is easy to apply in any Starcounter app.</p>
+            <p slot="kitchensink/main-description">Use the menu on the left to browse various examples of how to build web-based UI with Starcounter and Polymer.</p>
```
tomalec: I'd move above nodes before `<dom-bind>` to showcase that we don't need dom-bind unless we bind a model to it.

src/KitchenSink/wwwroot/sys/github-source-links/github-source-links.html
```diff
@@ -0,0 +1,74 @@
...
+    </style>
+    <div class="uni-section-primary">
+        <div class="uni-card github-source-links-links-container">            
+            <h3>See source code</h3>
+        </div>
+    </div>
+</template>
+<script>
+    const appVersion = document.querySelector('dom-bind').model.KitchenSink_0.appVersion;
+    debugger
```
tomalec: Please remove this `debugger`

src/KitchenSink/wwwroot/sys/github-source-links/github-source-links.html
```diff
@@ -0,0 +1,74 @@
...
+    } 
+    </style>
+    <div class="uni-section-primary">
+        <div class="uni-card github-source-links-links-container">            
+            <h3>See source code</h3>
+        </div>
+    </div>
+</template>
+<script>
+    const appVersion = document.querySelector('dom-bind').model.KitchenSink_0.appVersion;
```
tomalec: That way of accessing this version by relying on the presence of specific, separate view could be fragile, but I don't have any better idea.

src/KitchenSink/MasterPage.json
```diff
@@ -1,5 +1,7 @@
 {
   "Html": "/KitchenSink/MasterPage.html",
   "NavPage": {},
-  "CurrentPage": {}
+  "CurrentPage": {},
+  "appVersion": "",
```
PatrykSzwer: I don't really understand why you added those properties to MasterPage.json? They are needed only in MainPage.

src/KitchenSink/MainPage.json
```diff
@@ -1,3 +1,5 @@
 {
-  "Html": "/KitchenSink/MainPage.html"
+  "Html": "/KitchenSink/MainPage.html",
+  "appVersion": "",
```
PatrykSzwer: view-model property names should start with capital letter, so appVersion -> AppVersion etc.

src/KitchenSink/wwwroot/sys/github-source-links/github-source-links.html
```diff
@@ -0,0 +1,74 @@
...
+
+    class GTSL extends HTMLElement {
+        constructor() {
+            super();
+            this.attachShadow({ mode: 'open' });
+        }
+        guessCurrentPage() {
+            const pathname = window.location.pathname;
+            const lastPartOfPathName = pathname.substr(pathname.indexOf('/', 1) + 1);
+            const fullPageName = lastPartOfPathName.endsWith('Page') ? lastPartOfPathName : lastPartOfPathName + 'Page';
```
tomalec: This part emphasizes the need for KitcheSink file structure refactoring :)

src/KitchenSink/Program.cs
```diff
@@ -7,6 +7,16 @@ namespace KitchenSink
 {
     class Program
     {
+        static string getAppVersionFromAssemblyFile()
```
PatrykSzwer: Those methods shouldn't be here. You can move them directly in MainPage.json.cs. and use them in getters for view-model properties. This should be a much cleaner solution.


Also, you should write method names with first Capital letter.

src/KitchenSink/wwwroot/sys/github-source-links/github-source-links.html
```diff
@@ -0,0 +1,74 @@
...
+    debugger
+    const baseLink = `https://github.com/Starcounter/KitchenSink/blob/${appVersion}/src/KitchenSink/`;
+    const template = document.currentScript.previousElementSibling;
+
+    class GTSL extends HTMLElement {
+        constructor() {
+            super();
+            this.attachShadow({ mode: 'open' });
+        }
+        guessCurrentPage() {
```
tomalec: I'd put it outside of element's API, it does not even use `this`.

src/KitchenSink/Program.cs
```diff
@@ -17,7 +27,14 @@ static void Main()
 
             Handle.GET("/KitchenSink/json", () => new Json());
 
-            Handle.GET("/KitchenSink/partial/mainpage", () => new MainPage());
+            Handle.GET("/KitchenSink/partial/mainpage", () => {
+                return new MainPage
+                {
+                    starcounterVersion = getStarcounterVersion(),
```
PatrykSzwer: This is unnecessary, use getters instead.

src/KitchenSink/wwwroot/sys/github-source-links/github-source-links.html
```diff
@@ -0,0 +1,74 @@
...
+            const pathname = window.location.pathname;
+            const lastPartOfPathName = pathname.substr(pathname.indexOf('/', 1) + 1);
+            const fullPageName = lastPartOfPathName.endsWith('Page') ? lastPartOfPathName : lastPartOfPathName + 'Page';
+            return fullPageName;
+        }
+        createAnchor(text, url) {
```
tomalec: I'd put it outside of element's API

src/KitchenSink/wwwroot/sys/github-source-links/github-source-links.html
```diff
@@ -0,0 +1,74 @@
...
+            const anchor = document.createElement('a');
+            anchor.href = url;
+            anchor.innerHTML = text;
+            return anchor;
+        }
+        createElements() {
```
tomalec: I'd put it outside of element's API

src/KitchenSink/Program.cs
```diff
@@ -276,6 +293,8 @@ public static MasterPage GetMasterPageFromSession()
             {
                 master = new MasterPage
                 {
+                    appVersion = getAppVersionFromAssemblyFile(),
+                    starcounterVersion = getStarcounterVersion(),
```
PatrykSzwer: This is completely unnecessary, you can remove this part since you're not using those anywhere.

src/KitchenSink/MasterPage.json
tomalec: They are not only needed there. `github-source-links` uses it https://github.com/Starcounter/KitchenSink/pull/301/files/729609a2a7a44fd29b342764c35fb4dc411f31d9..45c056ca4ea7e6676e4ada408c3778a4758cec1d#diff-3dca39f6ede15cc8de5d16cf71c177ecR31

src/KitchenSink/Program.cs
tomalec: https://github.com/Starcounter/KitchenSink/pull/301/files/729609a2a7a44fd29b342764c35fb4dc411f31d9..45c056ca4ea7e6676e4ada408c3778a4758cec1d#diff-3dca39f6ede15cc8de5d16cf71c177ecR31

src/KitchenSink/MasterPage.json
PatrykSzwer: Ou, okey, still though as far as I see there is only appVersion needed, SC version is not.

PatrykSzwer: So I would suggest to left method of getting appVersion in Program.cs and get it for MasterPage and then i MainPage just use Parent data to update its own property. Something like this should work:

alshakero:
All comments are addressed not. I created an `/KitchenSink/kitchensink-app-version` endpoint to server the version.

src/KitchenSink/wwwroot/KitchenSink/MainPage.html
```diff
@@ -8,7 +8,7 @@ <h1 slot="kitchensink/main-heading">Welcome to KitchenSink</h1>
             <p slot="kitchensink/main-description">In this app you can take a look on typical UI patterns in a form that is easy to apply in any Starcounter app.</p>
             <p slot="kitchensink/main-description">Use the menu on the left to browse various examples of how to build web-based UI with Starcounter and Polymer.</p>
             <p slot="kitchensink/main-app-version">
-                <em>App version: {{model.appVersion}}, running on Starcounter: {{model.starcounterVersion}}</em>
+                <em>App version: <span data-app-version>{{model.appVersion}}</span>, running on Starcounter: <span data-sc-version>{{model.starcounterVersion}}</span></em>
```
tomalec: Is `data-*-version` a selector for tests only? Can't you use slot attribute selector for that?

src/KitchenSink/wwwroot/sys/github-source-links/github-source-links.html
```diff
@@ -1,73 +1,84 @@
...
+        <div class="uni-card github-source-links-links-container">
             <h3>See source code</h3>
         </div>
     </div>
 </template>
 <script>
-    const appVersion = document.querySelector('dom-bind').model.KitchenSink_0.appVersion;
-    debugger
-    const baseLink = `https://github.com/Starcounter/KitchenSink/blob/${appVersion}/src/KitchenSink/`;
+    async function getBaselink() {
```
tomalec: Could you make those functions local, not to pollute window scope?

src/KitchenSink/wwwroot/sys/github-source-links/github-source-links.html
```diff
@@ -1,73 +1,84 @@
...
-            card.appendChild(this.createAnchor('C# Code', csharp));
-            card.appendChild(this.createAnchor('JSON code-behind code', json));
-            card.appendChild(this.createAnchor('HTML view code', view));        
-
-            return content;
-        }
-        connectedCallback() {
-            this.shadowRoot.appendChild(this.createElements());
+        async connectedCallback() {
+            const baseLink = await getBaselink();
```
tomalec: I'd fetch it once for all instances, before any element is even stamped, not to send a request for each and every instance being (re-)stamped

alshakero: Did it like this in the start then thought `connectedCallback` is called once. Only forget the rest (`... per instance`) 🤣

tomalec:
Why does "design" pages don't have those links? I'd add them too.
see http://localhost:8080/KitchenSink/Alerts

Also, margins between cards are uneven, is that desired result? 
image

alshakero:
Regarding the spacing I can't think of an easy (non-hacky) fix. Sections have margins.

src/KitchenSink/wwwroot/KitchenSink/Design/NativePage.html
```diff
@@ -7,6 +7,8 @@ <h1 slot="kitchensink/native-heading">Native elements</h1>
             <p slot="kitchensink/native-description">Starcounter <a target="_blank" href="https://docs.starcounter.io/v/2.4/topic-guides/blendable-web-apps/app-shell">app shell</a> provides Underwear by default, <a target="_blank" href="https://starcounter.github.io/underwear.css/">here</a> you can check how every nativ element look like.</p>
```
tomalec: could you fixe `nativ` -> `native` typo?

src/KitchenSink/wwwroot/KitchenSink/Design/NativePage.html
```diff
@@ -7,6 +7,8 @@ <h1 slot="kitchensink/native-heading">Native elements</h1>
...
         </template>
     </dom-bind>
+    <github-source-links hide-csharp hide-json></github-source-links>
```
tomalec: Do you think this page needs a source code to be shown? On one hands it is consistent, but there is just a plain text :|

src/KitchenSink/wwwroot/KitchenSink/Design/LeftNavLayoutPage.html
```diff
@@ -21,6 +21,8 @@ <h1 slot="kitchensink/leftnavlayout-heading">Layout with left side navigation</h
             <label slot="kitchensink/leftnavlayout-navigation-content">Main content goes here</label>
         </template>
     </dom-bind>
+    <github-source-links hide-csharp hide-json></github-source-links>
```
tomalec: Could you remove line no.14 BTW?

alshakero:
All addressed

Issue #300, opened 2018-05-01
Pages with asynchronous behaviors stopped working


PatrykSzwer:
Related to https://github.com/joozek78/starcounter-async/issues/1

Every page that uses `AsyncInputHandlers` is not working now. Those pages are:
- Flash message
- Progress bar
- Lazy loading

After https://github.com/joozek78/starcounter-async/issues/1 is fixed this issue should be closed.

PatrykSzwer:
Closing as done. #306

PR #306, opened 2018-05-14
Update Async package

PatrykSzwer:
For #300


alshakero:
Can you please add tests for Async pages? Tests were passing even though those pages were broken.

PatrykSzwer:
Good idea, I'll handle that!

PatrykSzwer:
I added test for ProgressBar page that will fail if async library will stop working again. @alshakero please take a look.

PatrykSzwer:
Build succeed, merging. Thank you!

PR #309, opened 2018-05-17
Move styles from imported documents #282

tomalec:
### Done: 
- Move KitchenSink/style.css from imported document to imported view.
   As this is deprecated.
- Update paper elements

### Missing
- `font-roboto` is using `link rel="stylesheet` in imported document https://github.com/PolymerElements/font-roboto/issues/20,
- `vaadin-date-picker@2.x` is using styles from imported documents. `@3.0.0` fixes that, but is not stable enough, at least Material Theme does not work yet. Let's hope we will have it in following weeks https://github.com/vaadin/vaadin-material-theme/issues/35#issuecomment-389865254

Other styles are the ones used by Polymer's `<custom-style>` which moves `<style>` children to the main document, so should not make any problems once support will be dropped.


alshakero:
Forgot to mention. I checked every page and the only page still warning is `http://localhost:8080/KitchenSink/Radio` I assume it's because of Roboto.

alshakero:
Ah I noticed that some pages don't show the warning unless you fully load the page(as opposed to Palindrom-navigate). That's why I missed the ones you mentioned.

PR #317, opened 2018-06-05
Fix Github source code links

alshakero:
I now guess the page location using `model.Html` and things are much cleaner this way. Also, to feel confident about these links, given they're probably the first thing a prospective customer might see, I added tests for each page to make sure none of the links is 404. 

This means that the tests need internet to run now, which I think is alright. If you think not, I can make them pass when there is no internet but not when it's 404 in one extra line. I didn't do this because it's opinionated. Please share your thought. 

Fixes: https://github.com/Starcounter/KitchenSink/issues/313


src/KitchenSink/wwwroot/KitchenSink/Components/CheckboxPage.html
```diff
@@ -16,7 +16,7 @@ <h1 slot="kitchensink/checkbox-heading">Checkbox</h1>
             <input id="kitchensink-checkbox-input" slot="kitchensink/checkbox-input" type="checkbox" checked="{{model.DrivingLicense$::change}}">
 
             <div slot="kitchensink/checkbox-license-label">{{model.DrivingLicenseReaction}}</div>
-            <github-source-links></github-source-links>
+            <github-source-links base="{{model.Html}}"></github-source-links>
```
warpech: `github-source-links` is not imported in this and few more views

src/KitchenSink/wwwroot/KitchenSink/elements/github-source-links.html
```diff
@@ -31,25 +31,23 @@ <h3>See source code</h3>
...
-        async function getBaselink() {
-            if(cachedBaseLink) return cachedBaseLink;
+        let cachedGitHubBaseURI;
+        async function getGitHubBaseURI() {
+            if(cachedGitHubBaseURI) return cachedGitHubBaseURI;
             const appVersion = await fetch('/KitchenSink/kitchensink-app-version').then(res => res.text());
-            const baseLink = `https://github.com/Starcounter/KitchenSink/blob/${appVersion}/src/KitchenSink/`;
-            return (cachedBaseLink = baseLink);
+            const GitHubBaseURI = `https://github.com/Starcounter/KitchenSink/blob/${appVersion}/src/KitchenSink/`;
+            return (cachedGitHubBaseURI = GitHubBaseURI);
```
warpech: Is this code golf? I need to pause for 15 seconds on this line to understand what it does. Breaking it into 2 lines would save me that time.

src/KitchenSink/wwwroot/KitchenSink/elements/github-source-links.html
```diff
@@ -31,25 +31,23 @@ <h3>See source code</h3>
...
-        let cachedBaseLink;
-        async function getBaselink() {
-            if(cachedBaseLink) return cachedBaseLink;
+        let cachedGitHubBaseURI;
+        async function getGitHubBaseURI() {
+            if(cachedGitHubBaseURI) return cachedGitHubBaseURI;
             const appVersion = await fetch('/KitchenSink/kitchensink-app-version').then(res => res.text());
-            const baseLink = `https://github.com/Starcounter/KitchenSink/blob/${appVersion}/src/KitchenSink/`;
-            return (cachedBaseLink = baseLink);
+            const GitHubBaseURI = `https://github.com/Starcounter/KitchenSink/blob/${appVersion}/src/KitchenSink/`;
```
warpech: This is not even a URI but also a URL!

Rule of thumb: It is a URL if it has the protocol part. Otherwise it is a URI.

alshakero: Strange because the browser calls something very similar a URI. It was what inspried me 
image

tomalec: That one is tricky, MDN uses URL to define it https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI as well as WHATWG. But DOM-Level-3 uses URI and references to XML spec, I bet it's some kind of legacy "I" ;)

alshakero:
I won't release a version now. Maybe we'll adopt `uni-pagination` in KS tomorrow.

PR #323, opened 2018-06-13
Rename to UnifromDocs

alshakero:
I don't think this is easily reviewable. I replaced as such

`kitchensink` => `uniformdocs`.
`KitchenSink` => `UniformDocs`.
`Kitchensink` => `Uniformdocs`.

And did the neccessary house keeping (files etc) until all tests passed.

Naturally, GitHub source links will fail until we rename the repo.

https://github.com/Starcounter/KitchenSink/issues/319

**Edit: I renamed the repo to make sure we don't merge a failing test.**


alshakero:
Boiiiiiiiiiiiii
image

alshakero:
You are right, my comment is misleading. 

Yes it won't work because it points to the version tag and version 2.4.0 still has the name KitchenSink.

alshakero:
Should I give it version 3.0?

PatrykSzwer:
Thats actually a good idea, to bump major version.

PR #325, opened 2018-06-14
Add Data Table page

platosha:
Depends on Starcounter/uniform.css#123


PatrykSzwer:
I added some changes and improvements to the server-side myself.
About Components instead of Design I was thinking about it. My first though was that this element is part of the uniform, our design system and it should serve to improve the design, therefore should be in the Design section. 
But now we changed Kitchensink to UniformDocs itself, so maybe now it suites better in the Components section. @warpech maybe you will take deciding vote?

src/KitchenSink/ViewModels/Design/DataTablePage.json
```diff
@@ -2,7 +2,8 @@
   "Html": "/KitchenSink/Design/DataTablePage.html",
   "Columns": [
     {
-      "Header": "",
+      "Path": "",
```
tomalec: What this path refers to?

PatrykSzwer: We agreed with @platosha to add those for easier identification of columns during all server-side processes, like sorting or filtering. Path will be equal to the property of data model that column refers to. So if your data view-model has property `LastName`, `Path` for column of last names will also be `LastName`.

warpech:
> But now we changed Kitchensink to UniformDocs itself, so maybe now it suites better in the Components section. @warpech maybe you will take deciding vote?

Definitely "Components".

PatrykSzwer:
Ok, I'll move it.

src/KitchenSink/ViewModels/Design/DataTablePage.json
tomalec: What if my column renders:

PatrykSzwer:
Guys, the situation is that during the existence of this PR, we changed the entire structure of Kitchensink by renaming it to UniformDocs (related PR: https://github.com/Starcounter/UniformDocs/pull/323). That result in multiple conflicts that would be really hard and painfull to fix here. Instead I created new branch (https://github.com/Starcounter/UniformDocs/tree/uniformDocs-uni-data-table), based on current `master-2.4` and copied all changes made in this PR there. I asked @platosha to check it, and if everything will be fine, we will close this PR and create new one, which will continue current work.

platosha:
Closing in favour of #328

Issue #324, opened 2018-06-13
Be more selective with tests you give more than once chance


alshakero:
Currently I'm repeating failing tests 5 times to handle stale element references. If the tests is actually failing it's taking 50 seconds! That's a very long time. 

I will only retry when then exception is of StaleElementReference type.

warpech:
Moving this to "Backlog" to decide if the PR #330 should be closed or not.

PR #330, opened 2018-07-03
Try every test only once unless asked otherwise

alshakero:
See https://github.com/Starcounter/UniformDocs/issues/324

### DIFF
--- a/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
+++ b/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
@@ -63,7 +63,7 @@ namespace UniformDocs.Tests.Test
         [Test]
         public void AutoCompletePage_GitHubSourceURL()
         {
-            WaitUntil(x => _autoCompletePage.GitHubSourceLinks.Displayed);
+            WaitUntil(x => _autoCompletePage.GitHubSourceLinks.Displayed, "", 10, 3);
             TestGitHubSourceLinkURLs();
         }
     }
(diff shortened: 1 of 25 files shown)


alshakero:
@kegor ping.

alshakero:
I would love to do that and do it like all other tests. But then we'd have a lot of repeated code for the same element and the same expected behaviour. 

> This PR is a workaround for the previous workaround. We should not be building piles of workarounds.

This PR removes the workaround for all other tests except `GitHubSourceLinks`, I see it more like it mostly cancels the other workaround. And BTW, `waitUntil` natively has an option to ignore this exception, 
By doing: 
```csharp
wait.IgnoreExceptionTypes(typeof(NoSuchElementException), typeof(StaleElementReferenceException));
```
and it's a very common scenario. But this feature is broken in Selenium so I implemented my own exception ignoring setup. 

See https://github.com/SeleniumHQ/selenium/issues/4240

And commit history here, https://github.com/Starcounter/UniformDocs/pull/331#commits-pushed-1c24e7f

warpech:
This PR should be retargeted to `master` (which is equivalent to was was `master-2.4` until now) or a new PR should be made.

warpech:
@alshakero and @kegor WDYT of closing this PR without merging? I never fully understand the reason for this change, but I can tell that the tests of UniformDocs have not been flaky for very long time, so this PR might be not needed.

kegor:
From one side I agree with @warpech in the sense that if tests running ok on TC and locally then we don't have reasons to make any fixes, but from another side i didn't understand fully the case of running these tests in which @alshakero met the problem and what kind of problem. Maybe this problem isn't actual with using `BrowserStack` at nowdays?

alshakero:
I think this PR can be closed without merging.

Issue #318, opened 2018-06-12
Move text from Uniform and Underwear docs and demo pages to KitchenSink


warpech:
As discussed on the today's meeting between me and @alshakero, @PatrykSzwer, @tomalec and Rolf, we are merging docs of Underwear and Uniform into KitchenSink.

As the first step in that work, move text from Uniform and Underwear docs and demo pages to the relevant KitchenSink pages.

PR #328, opened 2018-06-29
Add DataTable page

platosha:
Reimplement all the changes from https://github.com/Starcounter/UniformDocs/pull/325 PR, due to conflicts after changing kitchensink to uniformDocs.

PR #331, opened 2018-07-05
Move uniform.css demos and use all available uni-components

alshakero:
Only `<uni-form-item-group>` is left to do. I'll do it tomorrow after Omni meeting.

Requires installing `uniform.css` from master to work. 

Fixes https://github.com/Starcounter/UniformDocs/issues/318

The issue is unclear. So I did all the obvious (and most time consuming) but work and some questions remain that need easy work. Will there then still be separate `uniform.css` docs? Or we're copying everything and removing them from there? This PR assumes yes. If not, I'd appreciate some document or comment that describes what goes where.


warpech:
> Will there then still be separate uniform.css docs?

There  will be no separate Uniform.css docs. But the `index.html` files can be kept for testing purposes.

alshakero:
I'm now adding an example of `<uni-form-item-group>` and after that it should be reviewable.

alshakero:
Should be ready. Although requires `uniform.css#0.4.0`.

warpech:
I have merged PR #328 into here. It includes an up-to-date demo of `uni-date-picker`.

alshakero:
Merging as [released ](https://github.com/Starcounter/UniformDocs/releases/tag/3.1.0)and [deployed](https://uniform.starcounter.io/).

PR #338, opened 2018-07-17
Replace implementation of uni-data-table server side with helping library

PatrykSzwer:
During implementation I found out that deleting rows won't work right now, due to https://github.com/Starcounter/Starcounter.Uniform/issues/7

> C# API
> Server-side implementation including population of data-model is provided by [Starcounter.Uniform library](https://github.com/Starcounter/Starcounter.Uniform/blob/master/README.md).


src/UniformDocs/ViewModels/Components/DataTableRow.json.cs
```diff
@@ -0,0 +1,13 @@
+using Starcounter;
+using UniformDocs.Database;
+
+namespace UniformDocs.ViewModels.Components
+{
+    partial class DataTableRow : Json, IBound<TableRow>
+    {
+        public void Handle(Input.DeleteTrigger action)
+        {
+            this.Data.Delete();
```
warpech: I have a good and bad news.

Good news is that I have a solution to the problem xxxx.

src/UniformDocs/Program.cs
```diff
@@ -1,12 +1,15 @@
-﻿using System;
+﻿using Starcounter;
+using Starcounter.Advanced;
```
warpech: `using Starcounter.Advanced;` is not used

src/UniformDocs/ViewModels/Components/DataTablePage.json.cs
```diff
@@ -1,20 +1,10 @@
...
 {
     partial class DataTablePage : Json
     {
-        public void Init()
-        {
-            this.DataTable.PopulateColumns();
-            this.DataTable.ReloadRows();
-        }
-
         void Handle(Input.AddNewRowTrigger action)
```
warpech: Clicking on `AddNewRowTrigger` has no effect until the page is reloaded. It should be reflected immediately in the table.

warpech:
Overall, the helper API looks brilliant to me! Great job @PatrykSzwer and @joozek78 💯!

I reported minor comments above.

@miyconst perhaps you'd like to review as well.

miyconst:
Currently I am busy with some other urgent tasks, but after the brief review it looks really promising. I will review the solution more deeply after my vacation.

src/UniformDocs/ViewModels/Components/DataTablePage.json.cs
```diff
@@ -28,181 +18,4 @@ void Handle(Input.AddNewRowTrigger action)
             });
```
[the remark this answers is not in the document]
PatrykSzwer: Sure, no prob

src/UniformDocs/Program.cs
PatrykSzwer: Yes it is, there are Blender.MapUri functions in Program.cs.

src/UniformDocs/Program.cs
```diff
@@ -149,8 +152,20 @@ private static void Main()
...
+                            .WithDataSource(DbLinq.Objects<TableRow>())
+                            .WithColumns(columns =>
+                                columns
+                                    .AddColumn(b => b.FirstName,
+                                        column => column.DisplayName("First Name").Sortable().Filterable())
+                                    .AddColumn(b => b.LastName, column => column.Sortable().DisplayName("Last Name"))
+                                    .AddColumn(b => b.Email,
+                                        column => column.Filterable().Sortable().DisplayName("Email")))
+                            .Build()
+                    };
```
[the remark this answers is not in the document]
PatrykSzwer: I would say that yes, It's less code and more clear to have `Builder` call inside a handler. If you insist to favorite source link I can move it to some `Init()` function to the view-model, but since UniformDocs shows our practices on how to do stuff people immediately will start repeating the same pattern.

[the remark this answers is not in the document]
PatrykSzwer: I talked with @joozek78 and I changed my mind. Due to the fact how Authorization library is implemented, and fact that in its case handlers are implemented through attributes, I agree that better way will be to include builder call inside view-model file instead of Program.cs -> handlers. I'll change that.

src/UniformDocs/ViewModels/Components/DataTablePage.json.cs
PatrykSzwer: Fixed.

src/UniformDocs/Program.cs
warpech: I see now.... 

It is not correct to have `using Starcounter.Advanced;` in any app other than BlendingProvider. Can you please replace `Blender.MapUri` calls with `blend.json`, like here: https://github.com/Starcounter/Blending/blob/master/src/BlendingEditor/blend.json

PatrykSzwer:
I added `blend.json` file. I'll merge this but please check anyway, and If something will be wrong, make a separate issue about it.

PR #340, opened 2018-07-20
Uni-form-item and uni-form-item-group with Starcounter.Uniform library

PatrykSzwer:
Small note: I removed initial invalid messages from inputs in form item group page, because it looks better for me and I think It's more often approach, to not have invalid messages right after entering the page.


src/UniformDocs/Program.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using Starcounter;
+using Starcounter.Advanced;
```
warpech: This `using` is not used. Please remove

src/UniformDocs/ViewModels/Components/FormItemGroupPage.json.cs
```diff
@@ -1,86 +1,71 @@
...
-            get
+            if (City.Length == 0 || Postcode.Length == 0)
             {
-                if (City.Length > 0 && Postcode.Length > 0)
-                {
-                    return City.ToLower().Equals("stockholm") && Postcode.Equals("12345") ? "false" : "true";
-                }
-
-                return "";
+                this.FormItemMetadata.SetMessage(GroupPropertyName, "Expecting 'Stockholm' and '12345'", MessageType.Neutral);
```
warpech: Here we mixed two coding styles. With `this.` and without it.

According to https://github.com/Starcounter/CompanyTrack/blob/master/AppsTeam/Guidelines/coding_style/csharp_code_formatting.md, `this.` should not be omitted. Could you pls fix my code to include it?

PatrykSzwer: Sure

src/UniformDocs/Program.cs
PatrykSzwer: Done

Issue #339, opened 2018-07-20
Document the data table component


warpech:
Merge the docs from:

- https://github.com/Starcounter/uniform.css/blob/master/components/uni-data-table/README.md - https://github.com/Starcounter/Starcounter.Uniform/blob/master/README.md 

into the page https://uniform.starcounter.io/UniformDocs/DataTable

warpech:
Closing as done in #341

PR #341, opened 2018-07-24
Uniform data table docs

warpech:
I am not fully happy with the page content, but I need to switch to other tasks.

In future, I would like to add the documentation of the C# API directly in UniformDocs, too.

Implements #339.


src/UniformDocs/wwwroot/UniformDocs/Components/DataTablePage.html
```diff
@@ -34,6 +36,129 @@ <h1 slot="uniformdocs/datatable-heading">Data Table</h1>
...
+            <p slot="uniformdocs/developer">The <code>&lt;uni-data-table-column&gt;</code> element expects one or more <code>&lt;template&gt;</code> elements as its light DOM, with the following purpose:</p>
+
+            <table slot="uniformdocs/developer">
+                <tr>
+                    <th>Slot</th>
+                    <th>Description</th>
+                </tr>
+                <tr>
+                    <td>(none)</td>
+                    <td>Template of the cell element. You can use it to provide your own renderer or editor for a cell. Available <code>dom-bind</code> properties in it: <code>index</code>, <code>item</code>.</td>
```
PatrykSzwer: This description is not really clear and it doesn't really describes what `index` and `item` are and what are they for. Please provide some more detailed information.

warpech:
I've added some clarifications, please re-review

src/UniformDocs/wwwroot/UniformDocs/Components/DataTablePage.html
```diff
@@ -97,7 +97,7 @@ <h4 slot="uniformdocs/developer">Column</h4>
                 </tr>
                 <tr>
                     <td>(none)</td>
-                    <td>Template of the cell element. You can use it to provide your own renderer or editor for a cell. Available <code>dom-bind</code> properties in it: <code>index</code>, <code>item</code>.</td>
+                    <td>Template of the cell element. You can use it to provide your own renderer or editor for a cell. It may utilize Polymer data-binding features as well as declarative event listeners to bind. The following properties are exposed to the template: <code>index</code> (the index of the row on the page), <code>item</code> (the row view-model). For example, use <code>&lt;template&gt;&lt;input value="{<!-- This comment is needed to prevent Polymer from parsing curly braces -->{item.FirstName$::change}}"&gt;&lt;/template&gt;</code> to have an input element that updates the server on each keystroke. Event listeners and inline computing functions in the template will be called on the host of the template, which usually is the closest <code>&lt;dom-bind&gt;</code> element up in the DOM tree.</td>
```
PatrykSzwer: One last thing that I would change here. You're writing description about template of the cell element, then you're explaining properties and then... you're giving example of usage without properties. I would either move example higher, right after

PR #347, opened 2018-07-31
Populate manifest dynamically

alshakero:
Note that the application tab in Chrome DevTools is not realtime. You have restart DevTools to see it populated.


warpech:
This is golden.

I am using this method in https://github.com/Starcounter/HeadsOmni/issues/11. 

I am not sure if this PR here should be accepted, perhaps not. It would be great to generalize this approach in Blending app suite on the Layouts level. There is a related issue https://github.com/Starcounter/Blending/issues/170

I will dig further.

alshakero:
Will close but keep the branch for inspiration.

Issue #348, opened 2018-08-01
Pressing Remove in uni-data-table page causes a server exception


alshakero:
It seems like the same bug as this https://github.com/Starcounter/Website/issues/176.

I don't fully understand what's going on, but commenting out [this line](https://github.com/Starcounter/UniformDocs/blob/master/src/UniformDocs/ViewModels/Components/DataTableRow.json.cs#L10) fixes the server side issue. Not sure if it is actually a solution though. 

Worth noting that there are two bugs here. Server and client bugs. This issue is meant to track only the server bug. In other words, if you get a `remove` patch after you click Remove button without an exception this issue can be considered fixed. Do not expect the row to be removed. The row removal bug is tracked in https://github.com/Starcounter/uniform.css/issues/158.

warpech:
I created a fix because @PatrykSzwer is on vacation. Fix is in PR #350

PR #350, opened 2018-08-02
initialize DataTable page in Db.Scope (#348)

warpech:
Because otherwise editing the email address or removing a row triggers an exception.

Fixes: https://github.com/Starcounter/UniformDocs/issues/348


alshakero:
I'm writing tests for uni-data-table to make sure my fixes work.

PR #357, opened 2018-08-14
Discussion PR: create ServiceWorker middleware

alshakero:
_This PR is not intended to be merged. It's just to use GH GUI to diff and discuss._

**I created a new middleware that does two jobs:** 
1. The same job as regular `PartialToStandaloneHtmlProvider`: delivers the app shell.
2. Handles requests to `/sw.js` and gives you a service worker that has a unique key within representing the cached files. This key changes when you restart the app, thus invalidating the service worker cache. 

**Caveat:**

You need to refresh the page to know that you have to invalidate the cache, by the time you know you invalidated it, you already consumed its files, you need yet another refresh. So it takes 2 refreshes after the app restart to get the fresh content.

**Possible solutions:**

Luckily, after the **first** refresh, an event is fired that new content is available in the service worker, so we could either programmatically reload the page there without waiting for user input, or we could display some UI like `palindrom-error-catcher` saying "New content available, please refresh". 

**How to try it:**
1. Run the app from `Add-service-worker-middleware` branch.
2. Go to http://localhost:8080/UniformDocs/
3. Restart the app.
4. Refresh the page and watch the console, you should see "New content is available; please refresh.".

Related to https://github.com/Starcounter/StarcounterClientFiles/pull/110


src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
```diff
@@ -0,0 +1,158 @@
...
+                    if (!string.IsNullOrEmpty(serviceWorkerTemplate))
+                    {
+                        template = serviceWorkerTemplate;
+                    }
+                    else
+                    {
+                        template = FetchServiceWorkerTemplate();
+                        serviceWorkerTemplate = template;
+                    }
+                    var SW = template.Replace("REPLACE_ME_WTH_RUNTIME_HASH", RUNTIME_CACHE_KEY);
```
alshakero: I didn't use `String.format` because it requires double brackets in the source file which is quite impractical. I even would like to change the `app-shell.html` one too if you agree.

src/UniformDocs/wwwroot/sys/app-shell/app-shell-with-service-worker.html
```diff
@@ -0,0 +1,74 @@
...
+                    }} else {{
+                      // At this point, everything has been precached.
+                      // It's the perfect time to display a
+                      // "Content is cached for offline use." message.
+                      console.log('Content is cached for offline use.');
+                    }}
+                  }}
+                }};
+              }};
+              window.addEventListener('reset-sw-cache', () => {{ debugger; registration.update() }} );
```
alshakero: This is a leftover from research phase. Please consider it not here.

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
```diff
@@ -0,0 +1,32 @@
+﻿<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>{0}</title>
+    <script>  
+    (function() {{
+      if ('serviceWorker' in navigator) {{
+          navigator.serviceWorker.getRegistrations().then(function(registrations) {{
```
alshakero: In case someone opts out of `ServiceWorkerMiddleWare.cs` the service worker would stick. This helps agains that.

warpech:
Using the provided steps to try it, I can't get it running (using Starcounter 2.4.990.127). I did that on a super-cleared cache in Chrome.

image

image

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
warpech: But I understand the pain, but we are interested in runtime performance. `String.format` appears to be faster than `String.replace`: https://stackoverflow.com/questions/3833044/what-would-be-faster-in-this-context-string-format-or-string-replace

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
```diff
@@ -0,0 +1,158 @@
...
+                    var response = new Response
+                    {
+                        BodyBytes = defaultEncoding.GetBytes(SW)
+                    };
+                    response.Headers["Content-Type"] = "text/javascript";
```
warpech: `application/javascript`

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
warpech: In Starcounter 2.4, middleware is global for all apps. `Application.Use` call in one app, affects all apps. Because of that, we must have all apps use **the same** middleware, otherwise the first registered middleware wins. And that depends only on the application starting order.

The only solution I can think of is to have `PartialJsonToStandaloneHtmlProvider` configurable through the database.

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                throw new Exception(@"Could not fetch /sys/app-shell/app-shell-with-service-worker.html");
+            }
+        }
+
+        private static string FetchServiceWorkerTemplate()
+        {
+            try
+            {
+                string serviceWorkerUrl = "/service-worker.js";
```
warpech: Seems that you forgot to commit this file

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            application.Use(MimeProvider.Html(this.Invoke));
+        }
+
+        private static string FetchAppShellTemplate()
+        {
+            try
+            {
+                string appShellHTMLUrl = "/sys/app-shell/app-shell-with-service-worker.html";
+                return Self.GET(appShellHTMLUrl).Body;
```
warpech: The app shell with a service worker should be fetched only if the original URL contains `https://`. Is there a way to know that?

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
alshakero: Damn. That's the reason behind the 500 error. I guess it worked for me because it was coming from SCF that I had running during development.

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
alshakero: But it works in localhost without HTTPS.

warpech: Ok so also check for the hostname?

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
warpech: Can you push it here? Or should I run SCF from the feature branch? I would slightly prefer to have it here.

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
alshakero: I'll find a way. Maybe it fails silently in HTTP which would be cool.

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
alshakero: Can't we make this new Middlware the default middleware? The need for unregistering is a need that I created by offering an opt-out. If we don't let developers opt out there will be no need for this.

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
alshakero: Done

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
alshakero: Yup. It fails silently in Chrome and console logs this in FF "Error during service worker registration: DOMException: "The operation is insecure.". Do we need to bother?

alshakero:
Ready for another testing round.

warpech:
It is impossible for me to test because I am still getting exactly the same error as described here: https://github.com/Starcounter/StarcounterClientFiles/pull/110#issuecomment-412497600

alshakero:
Can you try now please?

warpech:
Same problem :(

alshakero:
Can you show me the output of http://localhost:8080/sw.js?

warpech:
Actually your fix was correct, seems that Starcounter cached it for me until I've rebuilt the app. Really strange, but I am moving on with testing.

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
warpech: This line doesn't do anything, because `REPLACE_ME_WTH_RUNTIME_HASH` is not present in `service-worker.js`.

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
warpech: Maybe not for now.

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
```diff
@@ -0,0 +1,158 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace UniformDocs
+{
+    public class ServiceWorkerMiddleWare : IMiddleware
```
warpech: Please rename to `PartialToStandaloneHtmlProvider` due to my comment https://github.com/Starcounter/UniformDocs/pull/357#discussion_r210503153

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
warpech: Yes we can and we should. What I mean is that the name of the class must not be changed, it must be called `PartialJsonToStandaloneHtmlProvider`.

src/UniformDocs/wwwroot/service-worker.js
```diff
@@ -0,0 +1,104 @@
...
...js/entrypoints/custom-elements-es5-adapter-index.js","sys/webcomponentsjs/entrypoints/webcomponents-ce-index.js","sys/webcomponentsjs/entrypoints/webcomponents-hi-ce-index.js","sys/webcomponentsjs/entrypoints/webcomponents-hi-index.js","sys/webcomponentsjs/entrypoints/webcomponents-hi-sd-ce-index.js","sys/webcomponentsjs/entrypoints/webcomponents-hi-sd-ce-pf-index.js","sys/webcomponentsjs/entrypoints/webcomponents-hi-sd-index.js","sys/webcomponentsjs/entrypoints/webcomponents-sd-ce-index.js","sys/webcomponentsjs/entrypoints/webcomponents-sd-index.js","sys/webcomponentsjs/externs/webcomponents.js","sys/webcomponentsjs/gulpfile.js","sys/webcomponentsjs/src/post-polyfill.js","sys/webcomponentsjs/src/pre-polyfill.js","sys/webcomponentsjs/src/promise.js","sys/webcomponentsjs/src/unresolved.js","sys/webcomponentsjs/webcomponents-ce.js","sys/webcomponentsjs/webcomponents-hi-ce.js","sys/webcomponentsjs/webcomponents-hi-sd-ce.js","sys/webcomponentsjs/webcomponents-hi-sd.js","sys/webcomponentsjs/webcomponents-hi.js","sys/webcomponentsjs/webcomponents-lite.js","sys/webcomponentsjs/webcomponents-loader.js","sys/webcomponentsjs/webcomponents-sd-ce.js","sys/webcomponentsjs/webcomponents-sd.js"];
```
warpech: We discussed that everything should be in runtime cache, not in precache.

If you have so many resources in the precache, the first render is blocked until all of them are preloaded.

alshakero:
Ready for another round.

warpech:
@alshakero responded that there is some issue with the current version in Mac/Safari. He investigates using BrowserStack. If that fails, I will try on my Mac. I will see if we can get a real Mac for testing for @alshakero.

warpech:
I tried Mac/Safari 11.1.2 and the result is not that I have a blank screen, but I can see some console errors (only on the first run).

The information `Content is cached for offline use.` is missing.

screen shot 2018-09-03 at 11 52 01

warpech:
It seems that there are no specific problems with Safari anymore. I will re-review it fully tomorrow or on Wednesday.

Perhaps @tomalec would like to review, too?

alshakero:
To make reviewing easier, here is what to expect from a working setup:

1. Upon the first visit the app should act normal, but a log message should appear with the message "Content is cached for offline use".
2. Up the second visit, DevTools Network tap should have all asset requests responded by from service worker. Including `htmlmerger` urls. 
3. Now, navigate to another page, all assets that exist on the previous page should be servied from SW. Such as uniform.css, webcomponents-lite.js etc.
4. Restart the SC app, refresh the page, everything should come from the SW, although it's outdated content. This is undesired but expected behavior. After the page loads, you should see "New content available, please reload the page" in console. Refresh the page and all contect should come fresh from network.
5. Refresh once more and all content should come from SW.

All these refresh are soft refreshes. And `log` means Chrome DevTools log.

warpech:
Using `http://localhost:8080` in Chrome 69, on a clean cache, I get plenty of exceptions in console:

capture3

alshakero:
Looking into it

warpech:
I am observing discreptancies in how the service worker works compared to the expected behavior.

#### Steps to reproduce

I am doing steps 1-6 from my previous comment, end then restarting the UniformDocs app.

#### Expected

After the first refresh, I should see the `New content is available; please refresh` message. After the second refresh, I should see the updated file.

#### Actual

In Chrome, the new file is presented immediately, without the `New content is available; please refresh` message. This is wrong.

In Firefox, I get the "New content is available; please refresh." message. However, after the refresh I still get the old file. This is also wrong.

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
warpech: This is solved now.

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
```diff
@@ -0,0 +1,74 @@
...
+      // cached resources are updated in the background.
+
+      // To learn more about the benefits of this model, read https://goo.gl/KwvDNy.
+      // This link also includes instructions on opting out of this behavior.
+
+      (function() {{
+          function register() {{
+          if ('serviceWorker' in navigator) {{
+            window.addEventListener('load', () => {{
+              const swUrl = `/sw.js`;
```
warpech: Please use a more private and less obscure URI, such as: `/sys/service-worker.js`.

What is now `/sys/service-worker.js` can be renamed to `/sys/service-worker-source.js` or similar, with a comment that this file is not meant to be loaded directly but is used as a source file for dynamic generation of the service worker in the middleware.

alshakero: The idea is that the service worker url must be at least on the same level of URLs it wants to cache. Anything deeper will uncover some urls like `sc/htmlmerger`.

alshakero:
I can't repro the issue from https://github.com/Starcounter/UniformDocs/pull/357#issuecomment-419026877 in Chrome. 
sw-bug

Could you tell me more data? Like is it localhost, 127.0.0.1, or ngrok? What is the exact version of Chrome. And what is the exact version of SC.

warpech:
Problem https://github.com/Starcounter/UniformDocs/pull/357#issuecomment-419026877 is reproducible both with `localhost` and ngrok with Starcounter 2.4.990.127. Screen recording:

search2

alshakero:
Can you please try with DevTools closed?

alshakero:
Comment https://github.com/Starcounter/UniformDocs/pull/357#issuecomment-419026877 is resolved as it turns out it's a DevTools issue. 

Regarding comment https://github.com/Starcounter/UniformDocs/pull/357#issuecomment-419035331 I'm getting the same as FF behavior in your comment. The file is stuck even after file restart. I'm investigating

alshakero:
Apparently, the service-worker itself is being cached. But i'm not sure if its being cache within itself or within the normal browser cache.

[According to the spec](https://stackoverflow.com/a/40233462/1247990), the service worker does not cache itself, however the browser cache does and the service must instruct the browser using the classic `Cache-control` header. 

SC is already responding with `Cache-control: no-store`. But the service worker is being cached for sure.

Tomorrow I'll further investigate how to prevent this caching.

warpech:
AFAIK, service worker is *always* cached using the algorithm explained in the spec. It is updated in the background and activated upon next page refresh.

- https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers
- https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle

alshakero:
I think we're talking about two different things. I'm talking about the request to the `/service-worker.js` file itself.

warpech:
Me too. Am I wrong?

---

Last week, we discussed that we have no other way than accept how browser updates service worker. If a developer/tester wants it force refresh the static files, he will have the choice of:

- clean browser cache with Ctrl+Shift+Delete
- bypass service worker using Chrome DevTools > Applications > Service Worker
- restart any Starcounter app and double refresh the page
- call a URI that regenerates the service worker token and double refresh the page (@alshakero will develop this URI handler in middleware)

@alshakero will write a blog post about it.

@alshakero can you pls confirm all above is correct?

warpech:
Maybe through use of a service worker framework https://developers.google.com/web/tools/workbox/modules/, we would remove potential problems caused by our fresh knowledge of the Service Workers topic?

tomalec:
@alshakero speaking of problems with old `service-workers.js` being served even though new version of `-source` is ready, seems to this is a middleware that caches it https://github.com/Starcounter/UniformDocs/pull/357/files#diff-3b03ff4ce71ddd37f4282c36e6955245R40

tomalec:
@warpech @alshakero Please, correct me if I'm wrong. 

AFAIR we have only one blocking issue is: when the `service-worker-source.js` is modified new `service-workers.js` file is not taking action after two reloads.

I see two problems there:
1. According to spec, new service worker will not be activated until all clients are disconnected/closed/navigated away. So "New content is available; **please refresh**." is not enough to reload new SW (it's enough to load new cached resources).  As same worker serves all the tabs. [developers.google](https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle#updates)
     > Once successfully installed, the updated worker will wait until the existing worker is controlling zero clients. (Note that clients overlap during a refresh.)
2. Our middleware caches the service worker template for a run time of an application, and serves the old one. [see source](https://github.com/Starcounter/UniformDocs/pull/357/files#diff-3b03ff4ce71ddd37f4282c36e6955245R40)

I can imagine three solutions for each:

1. Activation
   1. To activate new SW change the message to be  "New service worker/app-shell/application/TBD is available. **Please close or navigate away from all tabs that are using this page and come back again.** That would be hard to do for non-developers. The question is how often end-users will close all tabs that uses Starcounter solution.
   2. Use `self.skipWaiting()` which should activate the new worker once it's installed. The downside is that we may get inconsistent result as some content will be already loaded by old worker, then the new requests will be handled by the new one. (It's not necesarily previous version, it may be realyyy old one.
   3. Use `self.skipWaiting()` and force reload the page. That will serve consisten version but, may lead to data loss, and as Jake Archibald said:
       > Users actively dislike data loss. It causes them great sadness.
2. Middleware
   1. Require app restart.
   2. Do not cache it there,
   3.  Forward internal `Self.GET` with `If-Modified-After` header or smth.

Do I miss something? Do you have any preference over solution? Could you confirm that `service-workers.js` file is update after the change, app restart, and closing/refreshing **all** controlled tabs?

tomalec:
> Because all these people will be developers. They don't have data to lose. But they do have finite patience when they get outdate HTML and JS.

I'd say developers could handle that with few simple steps
https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle#devtools
And why do you think "all these people will be developers"?

Personally, I'd go for 1.i or 1.ii and 2.ii.

@warpech what's your opinion?

warpech:
Can we do the same solution as https://github.com/deanhume/pwa-update-available (which is `1.i` with a user option to perform `1.iii`).

The message could be:

> A new version of this app is available [Refresh]

Any downsides to this?

tomalec:
> Can we do the same solution as deanhume/pwa-update-available (which is `1.i` with a user option to perform `1.iii`).

:+1:

> Any downsides to this?

The only thing I imagine is that we will have to implement Starcounter-specific, enduser-facing UI which will be hard to blend/change inthe app-shell (without overwriting some static files. (Or maybe we could have it as a part of root composition..)

Anyway. I'd make this toas/message a separate issue. And go with what we have here. It would also give us some feedback from early adopters.

alshakero:
> And go with what we have here

I don't think the current phase is workable. Files stick and you need to clear service worker cache. Whenever I switch to this PR, then switch back to something else, I'm affected by SW cache still working.

warpech:
> Because the app will update in front of the developer 99% of the time and rarely for users (upon each delpolyment). And when the app updated for users it's right after they land on the app. Not after they fill a couple forms.

This is not true. If our customer implements a continous integration scenario, the deployment might happen once (or more) per day. All end users will be affected by the service worker upgrade at that point.

> The only thing I imagine is that we will have to implement Starcounter-specific, enduser-facing UI which will be hard to blend/change inthe app-shell (without overwriting some static files. (Or maybe we could have it as a part of root composition..)

This is not the only one. We will also have "Save Bar". My idea is that such UI should be overridable from a layout.

tomalec:
Sounds good to me, just the fact `palindrom-error-catcher` will be not 100% just about Palindrom, is something that bothers me, but for sure we could ship it in same way.

alshakero:
OK. So I found something that seems too good to be true, and I need your help to verify that it's working. It's called [Immediate Claim Recipe - General Usage - ServiceWorker Cookbook](https://serviceworke.rs/immediate-claim_service-worker_doc.html). 

It's awesome. I implemented a variation of what they advise and it seems that the ServiceWorker is flushing the old cache *before* the page is rendered when there is a new version. Rendering the aforementioned toast needless. And all is working seemeless without bother the user or the developer. 

How to veryify:

1. Pull. 
2. Open UniformDocs.sln.
3. Go to http://localhost:8080/UniformDocs
4. Modify MainPage.html 
5. Refresh the page. You should see the outdated version.
6. Restart the app. 
7. Refresh the page, you should see the fresh version immediately! 😍 
8. To verify once more: edit MainPage.html once more but don't restart the app.
9. Refresh, you should get the outdated version.
10. Restart the app and refresh and you should see a fresh version.

tomalec:
I suspect it will activate new SW on already loaded pages in other tabs (`1.ii`), but will double check.

alshakero:
@tomalec I think we agree on the `window.reload` on new content idea. Right? Is there anything else blocking this? 

Would you like me to continue?

tomalec:
If you don't have anything else in Omni, feel free to take it. I was about to take it after lower hanging fruits in my backlog. 

To be honest, I'm not sure on what we agreed, as the idea starts to rotten in my head, after picking and dropping this topic few times ;) 

I'm fine with starting with `window.reload`, but I'm not sure how it will interact with unsaved changes warning. Plus, Marcin had a point about using existing solutions, not to learn the hard way on our own.

I start to think we should really use some existing solution and ship something sooner to start fiddling with a real rather hypothetical cases.

src/UniformDocs/Helpers/PartialJsonToStandaloneHtmlProvider.cs
```diff
@@ -0,0 +1,154 @@
+﻿using Starcounter;
+using Starcounter.XSON.Advanced;
+using System;
+using System.Text;
+
+namespace UniformDocs
+{
+    public class PartialJsonToStandaloneHtmlProvider : IMiddleware
```
warpech: This is not a correct name. Should be: `PartialToStandaloneHtmlProvider`

warpech:
Current status: waiting for https://github.com/Starcounter/Home/issues/439

src/UniformDocs/Helpers/PartialToStandaloneHtmlProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <summary>
+        /// Creates a new instance of <see cref="PartialToStandaloneHtmlProvider"/>
+        /// using the given standalone page template.
+        /// </summary>
+        public PartialToStandaloneHtmlProvider(string standaloneTemplate)
+        {
+            if (string.IsNullOrEmpty(standaloneTemplate)) throw new ArgumentNullException("standaloneTemplate");
+            appShell = standaloneTemplate;
+
+            Starcounter.Internal.AppsBootstrapper.WatchResources(ResourceChanged); //this will be called n times, where n is the number of Palindrom sessions. Move it somewhere else where it is called once per resource change event
```
warpech: I have tried to use the new API as explained in https://github.com/starcounter/home/issues/439#issuecomment-477173174.

However, the `ResourceChanged` action is never called on my machine. Alexey could you pls check?

alemoi: I just moved `Starcounter.Internal.AppsBootstrapper.WatchResources(PartialToStandaloneHtmlProvider.ResourceChanged);` to the end of `Main()` of UniformDocs
and not from `PartialToStandaloneHtmlProvider(string standaloneTemplate)` which is actually never called for me
to test i modified file `UniformDocs\src\UniformDocs\wwwroot\UniformDocs\AppIcon.html`

warpech: > `PartialToStandaloneHtmlProvider(string standaloneTemplate)` which is actually never called for me

Thanks, this guided me to the problem! I only called `WatchResources` from one of the two constructor overloads in the `PartialToStandaloneHtmlProvider` class.

warpech:
Good news! The `PartialToStandaloneHtmlProvider` middleware is now able to regenerate the cache key when a file changes on disk. In result, `service-worker.js` is now capable of getting its cache invalidated by the server. This requires Starcounter 2.4.990.267

@alshakero, can you please resume to this PR. Please go through the comments to check if everything is implemented. 

I recall we agreed to refresh the page when the cache key is invalidated, right?

src/UniformDocs/Helpers/ServiceWorkerMiddleWare.cs
warpech: I agree to keep `String.replace` for now.

alshakero:
> I recall we agreed to refresh the page when the cache key is invalidated, right? 

Exactly! I'll continue with this.

alshakero:
OK so this is ready for testing. All you need to do is run the app from this branch. What you might want to test is:

1. Make a C# change, restart the app, it should be reflected right away in the browser.
2. Make an HTML view change, it should be reflected right away in the browser, without restarting the app. 

By right away I mean after a 500ms blink.

/cc @warpech @mtornwall

warpech:
I will test it on Tuesday.

@alshakero, do you have an idea, how can a solution owner disable a Service Worker that was already installed on the clients? We need to implement some safety vents for times when an unwanted content gets installed and cached on the clients.

src/UniformDocs/wwwroot/sys/service-worker-source.js
```diff
@@ -0,0 +1,106 @@
+const version = 'REPLACE_ME_WTH_RUNTIME_HASH';
```
warpech: Is it possible to add a `console.info` with a confirmation that a Service Worker is currently active and static resources are served from the SW cache? I want to add a link to a Starcounter docs page that explains quirks of SW for developers who are not familiar with it. 

For the beginning, please give a link to the GitHub issue https://github.com/Starcounter/StarcounterClientFiles/issues/111. Let's gather the relevant information in that issue and I will later move it to Starcounter Docs.

src/UniformDocs/Helpers/PartialToStandaloneHtmlProvider.cs
```diff
@@ -0,0 +1,152 @@
...
+            {
+                throw new Exception(@"Could not fetch /sys/app-shell/app-shell.html");
+            }
+        }
+
+        private static string FetchServiceWorkerTemplate()
+        {
+            try
+            {
+                string serviceWorkerUrl = "/sys/service-worker-source.js";
```
warpech: Since this is part of the app shell, I think it makes sense to change the URI to: `/sys/app-shell/service-worker-source.js`

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
```diff
@@ -0,0 +1,77 @@
...
+
+        To learn more about the benefits of this model, read https://goo.gl/KwvDNy.
+        This link also includes instructions on opting out of this behavior.
+      */
+
+      (function() {{
+          function register() {{
+          if ('serviceWorker' in navigator) {{
+            window.addEventListener('load', () => {{
+              const swUrl = `/service-worker.js`;
```
warpech: Since this is part of the app shell, I think it makes sense to change the URI to: `/sys/app-shell/service-worker.js`

alshakero:
> @alshakero, do you have an idea, how can a solution owner disable a Service Worker that was already installed on the clients? We need to implement some safety vents for times when an unwanted content gets installed and cached on the clients.

This requirement means running this JS:

```js
navigator.serviceWorker.getRegistrations().then(registrations => {
    for(const registration of registrations) {
        registration.unregister()
    }
})
```

Possible solutions:

1. Add a button in DevTools.
2. We can serve a URI that includes this JS. 

It's worth noting that we can detect hard reloads using HTTP request headers. See this diff between normal vs hard reload. We might be able to make some decision on the server when we see such headers. https://www.diffchecker.com/N7NIbdve

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
warpech: I have just re-read https://github.com/Starcounter/UniformDocs/pull/357#discussion_r215575992 and now I member why this URI must be top-level.

alshakero: Service workers can only cache URLs that are either in the same or in the deeper level. 

`/app/sw.js` cannot cache `/image.jpg`. See https://stackoverflow.com/a/35780776/1247990

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
```diff
@@ -0,0 +1,74 @@
...
+              registration.onupdatefound = () => {{
+                const installingWorker = registration.installing;
+                installingWorker.onstatechange = () => {{
+                  if (installingWorker.state === 'installed') {{
+                    if (navigator.serviceWorker.controller) {{
+                      // At this point, the old content will have been purged and
+                      // the fresh content will have been added to the cache.
+                      // It's the perfect time to display a "New content is
+                      // available; please refresh." message in your web app.
+                      console.log('New content is available; please refresh.');
```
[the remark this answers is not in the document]
alshakero: This comment is no longer relevant as we're reloading the page right away.

alshakero:
Address all.

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
```diff
@@ -0,0 +1,81 @@
...
+                      // It's the perfect time to display a
+                      // "Content is cached for offline use." message.
+                      console.info('Content is cached for offline use.');
+                    }}
+                  }}
+                }}
+              }}
+            }}
+            catch(error) {{
+              console.error('Error during service worker registration:', error);
```
warpech: I think part of what @tomalec meant in his comment https://github.com/Starcounter/UniformDocs/pull/357#discussion_r218397417 was to add some indication about the source of the console messages coming from this file. Otherwise it might be hard to understand their origin. I would prepend the text in lines 47, 53, 60 with `[Starcounter App Shell]`

src/UniformDocs/Helpers/PartialToStandaloneHtmlProvider.cs
```diff
@@ -8,7 +8,7 @@ namespace UniformDocs
     public class PartialToStandaloneHtmlProvider : IMiddleware
     {
         static Encoding defaultEncoding = Encoding.UTF8;
```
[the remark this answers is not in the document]
alshakero: Yes it is for the app-shell template.

src/UniformDocs/wwwroot/sys/app-shell/service-worker-source.js
```diff
@@ -0,0 +1,127 @@
...
+
+/** logs a nice distinct console message */
+function niceLog() {
+  const args = Array.from(arguments);
+  const formattedMessages = args.reduce((acc, message) => {
+    acc.push('%c[SW]: ' + message);
+    acc.push(serviceWorkerFormatting);
+    return acc;
+  }, []);
+  console.info(...formattedMessages);
```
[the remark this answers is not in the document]
alshakero: > But I guess we can keep it for its novelty

exactly. I want developers to be able to skip it when looking for other messages easily.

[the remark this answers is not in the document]
alshakero: Changed for a more lightness-neutral (the L in HSL) greenish color

image

src/UniformDocs/Helpers/PartialToStandaloneHtmlProvider.cs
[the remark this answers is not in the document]
warpech: Sorry, I thought I searched the file but apparently not

src/UniformDocs/Helpers/ServiceWorkerBodyBytesSingleton.cs
```diff
@@ -68,10 +63,19 @@ public byte[] GetServiceWorkerBodyBytes()
         /// in case of delete, newUri is null
         /// in case of a new file, oldUri is null
         /// </summary>
-        /// <param name="resourceChanged"></param>
-        private void ResourceChanged(string newUri, string oldUri)
+        /// <param name="_resourceChanged"></param>
+        private void _resourceChanged(string newUri, string oldUri)
```
warpech: The convention explained in https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md only applies to fields, not properties nor methods. It should be renamed back to `ResourceChanged`

src/UniformDocs/Helpers/ServiceWorkerBodyBytesSingleton.cs
```diff
@@ -68,10 +63,19 @@ public byte[] GetServiceWorkerBodyBytes()
...
         {
-            RegenerateKeyAndHydrateServiceWorkerTemplate();
+            /* FIXME: currently this is throwing "You are trying to perform a Self call while not being on Starcounter scheduler"
```
warpech: Try the solution from https://docs.starcounter.io/topic-guides/transactions/running-background-jobs

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
```diff
@@ -0,0 +1,67 @@
+﻿<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>{0}</title>
+    <script>  
+      const logPrefix = '[Starcounter App Shell];'
```
warpech: This should be inside the IIFE to not pollute the global scope.

Since we are using ES6, you can also use a code block `{ }` instead of an IFFE.

src/UniformDocs/wwwroot/sys/app-shell/app-shell.html
```diff
@@ -30,20 +30,20 @@
                       // At this point, the old content will have been purged and
                       // the fresh content will have been added to the cache.
                       // It's the perfect time to reload the page.
-                      console.info(`${logPrefix} Detected a new service worker cache version, reloading...`)
```
warpech: Woops. I missed the curly braces problem in my review 😳

src/UniformDocs/Helpers/ServiceWorkerBodyBytesSingleton.cs
```diff
@@ -63,19 +60,23 @@ public byte[] GetBodyBytes()
...
+        private void ResourceChanged(string newUri, string oldUri)
         {
-            /* FIXME: currently this is throwing "You are trying to perform a Self call while not being on Starcounter scheduler" 
-            
             // the service worker source file has changed. Re-fetch the source template
             // this cancels the need for restarting the app(s) to get the effects of updating the service-worker-source template
-            if(newUri.EndsWith("service-worker-source.js"))
+            if (newUri.EndsWith("service-worker-source.js"))
+            {
+                ThreadPool.QueueUserWorkItem(async o =>
```
warpech: I think it is enough to use `Scheduling.RunTask`, without wrapping in `ThreadPool.QueueUserWorkItem`

warpech:
All review comments seem to be addressed, thanks @alshakero! The last build failed due to the full queue at BrowserStack. I have restarted the build.

warpech:
The build passed with success!

Note: This PR will not be merged, but has become the protoplast of two final PRs: 

- https://github.com/Starcounter/level1/pull/4929
- https://github.com/Starcounter/StarcounterClientFiles/pull/138

alshakero:
> protoplast 

Your vocabulary never seizes to blow me away.

warpech:
Closing, as this PR served its purpose. The inclusion of Service Worker in Starcounter is tracked in the following PRs:

- Starcounter/level1#4929
- Starcounter/StarcounterClientFiles#138

PR #358, opened 2018-08-15
Main, UI Kit, Section, Card page content improvements

warpech:
See the commit messages for the explanation of the changes.

There are minor improvements to other pages, too.


warpech:
Done, thanks for the idea. It looks way cleaner now

PR #377, opened 2018-09-21
Cloud testing with BrowserStack

warpech:
Initial implementation of using BrowserStack as the default Selenium server

See CONTRIBUTING.md for explanation

In WebDriverManager, I reverted `ChromeOptions` (etc) to `DesiredCapabilities`, because that seems to be the only API compatible with BrowserStack

Implements: https://github.com/Starcounter/RebelsLounge/issues/210; Step towards: https://github.com/Starcounter/AdminTrack/issues/578


warpech:
Current status: 11 tests are failing in BrowserStack

warpech:
I fixed one failing test and investigated few others:

- FileUploadPageTest(Firefox).FileUploadPage_UploadAFile
- FileUploadPageTest(Firefox).FileUploadPage_UploadAndDeleteAFile
   - there is an `ElementNotInteractableException` error due to `input[type=file]` being in Shadow DOM, emulated by polyfill. I am not sure if we can trick Selenium into thinking that this element is interactable. Perhaps yes, by moving the input to light DOM?
- UrlPageTest(Chrome).UrlPage_ClickLinkWithDownloadAttribute 
- UrlPageTest(Firefox).UrlPage_ClickLinkWithDownloadAttribute
   - no wonder these tests fail. The download feature works, but the test C# code expects the file to be downloaded at the same computer where the test runs. With remove Selenium server this is clearly not the case. We need a different way to checking if download was started, for example by creating a special download handler

Still investigating:

- TextPageTest(Chrome).TextPage_TextPropagationForPaperTextOnUnfocus
- TextareaPageTest(Chrome).TextareaPage_WriteToTextArea
- UrlPageTest(Firefox).UrlPage_ClickBlankTargettedLink

warpech:
@tomalec it is ready for review. The build passed: https://teamcity.starcounter.org/viewLog.html?buildId=69549&tab=buildResultsDiv&buildTypeId=Apps_UniformDocs_UniformDocsMaster24branchWithStarcounterDevelopNightly

I will run another build as an extra check for flakyness.

test/UniformDocs.Tests/Ui/FileUploadPage.cs
```diff
@@ -22,7 +22,10 @@ public FileUploadPage(IWebDriver driver) : base(driver)
 
         public void UploadAFile(string filePath)
         {
-            GetFileElement().SendKeys(filePath);
+            var inputFileElement = GetFileElement();
+            var script = "return arguments[0].classList.contains('style-scope') ? arguments[0].style.display = 'initial' : true;"; //in Shadow DOM polyfill (Firefox), Selenium cannot type to an input that is "display: none" and throws ElementNotInteractableException. This line fixes the problem
```
tomalec: Why are we using the input hidden in ShadowDOM instead of light dom counterpart exposed to the end user?

test/UniformDocs.Tests/Utilities/WebDriverManager.cs
```diff
@@ -1,46 +1,135 @@
...
-                        profile.SetPreference("browser.helperApps.neverAsk.saveToDisk", "image/svg+xml");
-
-                        driver = new RemoteWebDriver(remoteWebDriverUri, profile);
-
+                        var firefoxOptions = new FirefoxOptions();
+                        firefoxOptions.SetPreference("browser.download.folderList", 0); //0 is recommended per BrowserStack docs; https://www.browserstack.com/automate/c-sharp#enhancements-uploads-downloads
+                        firefoxOptions.SetPreference("browser.download.manager.focusWhenStarting", false);
+                        firefoxOptions.SetPreference("browser.download.useDownloadDir", true);
+                        firefoxOptions.SetPreference("browser.helperApps.alwaysAsk.force", false);
+                        firefoxOptions.SetPreference("browser.download.manager.alertOnEXEOpen", false);
```
tomalec: Why do we need this setting?

test/UniformDocs.Tests/Utilities/WebDriverManager.cs
```diff
@@ -1,46 +1,135 @@
...
+                capability.SetCapability("browserstack.user", Config.BrowserstackUsername);
+                capability.SetCapability("browserstack.key", Config.BrowserstackAccessKey);
+                capability.SetCapability("browserstack.debug", "true");
+                capability.SetCapability("browserstack.local", "true");
+                capability.SetCapability("browserstack.localIdentifier", Config.BrowserstackLocalIdentifier);
+                capability.SetCapability("browserstack.video", "false"); //video recording increases test time, see https://www.browserstack.com/automate/c-sharp
+                capability.SetCapability("browserstack.use_w3c", "true"); //needed for closing of window in UrlPage_ClickBlankTargettedLink for Firefox on BrowserStack; see https://github.com/SeleniumHQ/selenium/issues/5064
+            }
+
+            capability.SetCapability("project", "UniformDocs");
```
tomalec: Can we read app name from somewhere, not to hard-code it, So it will not get copy'n'pasted to other apps? ;)

warpech:
One new flaky test found: `PaginationPageTest(Firefox).PaginationPage_LastButton_GoesToLastPage `. I am looking into it.

https://teamcity.starcounter.org/viewLog.html?buildId=69552&tab=buildResultsDiv&buildTypeId=Apps_UniformDocs_UniformDocsMaster24branchWithStarcounterDevelopNightly#testNameId-7165210323739111639

test/UniformDocs.Tests/Utilities/WebDriverManager.cs
warpech: This is all pasted from the BrowserStack docs, link to which is few lines above

test/UniformDocs.Tests/Ui/FileUploadPage.cs
warpech: Because `starcounter-upload` was designed 2 years ago without consideration for separation of content and presentation. We should change that at some point

tomalec: I didn't meant any separation of presentation and content, but just mimicking user actions closely. So click on the actual element that end user clicks. User is able to click visual (not `display:none;`) control = "Select file" button - to submit a file. If user is able to do so, why Selenium isn't?



But, now I see that https://github.com/Starcounter/UniformDocs/blob/master/src/UniformDocs/wwwroot/UniformDocs/HowTo/FileUploadPage.html#L130-L146 `starcounter-include` is not self-sufficient element, and there is lot's of work around it. Therefore fixing that should be a separate issue and PR.

warpech:
It seems to me that it is not the test that was flaky, but the connection between BrowserStackLocal and BrowserStack.

I will merge this PR, let's see if the same problem as https://teamcity.starcounter.org/viewLog.html?buildId=69552&tab=buildResultsDiv&buildTypeId=Apps_UniformDocs_UniformDocsMaster24branchWithStarcounterDevelopNightly#testNameId-7165210323739111639 happens again.

PR #384, opened 2018-11-13
Add mobile support tests - recommended by StatCounter

alshakero:
This PR covers every Android and iOS version mentioned that fall under the categories in this [comment](https://github.com/Starcounter/HeadsOmni/issues/663#issuecomment-439863998).

### DIFF
--- a/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
+++ b/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
@@ -7,6 +7,17 @@ namespace UniformDocs.Tests.Test
     [TestFixture(Config.Browser.Chrome)]
     [TestFixture(Config.Browser.Edge)]
     [TestFixture(Config.Browser.Firefox)]
+    [TestFixture(Config.Browser.iOS11point2)]
+    [TestFixture(Config.Browser.Android7Nougat)]
+    [TestFixture(Config.Browser.Android8Oreo)]
+    [TestFixture(Config.Browser.iOS11point4)]
+    [TestFixture(Config.Browser.iOS12)]
+    [TestFixture(Config.Browser.iOS11point3)]
+    [TestFixture(Config.Browser.Android7Nougat)]
+    [TestFixture(Config.Browser.Android6Marshmello)]
+    [TestFixture(Config.Browser.Android8point1Oreo)]
+    [TestFixture(Config.Browser.Android7point1Nougat)]
+    [TestFixture(Config.Browser.Android5point1Lolliopop)]
     class AutoCompletePageTest : BaseTest
     {
         private AutoCompletePage _autoCompletePage;
--- a/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
+++ b/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
@@ -7,13 +7,13 @@ namespace UniformDocs.Tests.Test
     [TestFixture(Config.Browser.Chrome)]
     [TestFixture(Config.Browser.Edge)]
     [TestFixture(Config.Browser.Firefox)]
-    [TestFixture(Config.Browser.iOS11point2)]
-    [TestFixture(Config.Browser.Android7Nougat)]
+    [TestFixture(Config.Browser.iOS11_2)]
+    [TestFixture(Config.Browser.Android7)]
     [TestFixture(Config.Browser.iOS12)]
-    [TestFixture(Config.Browser.iOS11point3)]
-    [TestFixture(Config.Browser.Android6Marshmello)]
-    [TestFixture(Config.Browser.Android7point1Nougat)]
-    [TestFixture(Config.Browser.Android5point1Lolliopop)]
+    [TestFixture(Config.Browser.iOS11_3)]
+    [TestFixture(Config.Browser.Android6)]
+    [TestFixture(Config.Browser.Android7_1)]
+    [TestFixture(Config.Browser.Android5_1)]
     class AutoCompletePageTest : BaseTest
     {
         private AutoCompletePage _autoCompletePage;
--- a/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
+++ b/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
@@ -13,7 +13,6 @@ namespace UniformDocs.Tests.Test
     [TestFixture(Config.Browser.iOS11point4)]
     [TestFixture(Config.Browser.iOS12)]
     [TestFixture(Config.Browser.iOS11point3)]
-    [TestFixture(Config.Browser.Android7Nougat)]
     [TestFixture(Config.Browser.Android6Marshmello)]
     [TestFixture(Config.Browser.Android8point1Oreo)]
     [TestFixture(Config.Browser.Android7point1Nougat)]
--- a/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
+++ b/test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
@@ -9,12 +9,9 @@ namespace UniformDocs.Tests.Test
     [TestFixture(Config.Browser.Firefox)]
     [TestFixture(Config.Browser.iOS11point2)]
     [TestFixture(Config.Browser.Android7Nougat)]
-    [TestFixture(Config.Browser.Android8Oreo)]
-    [TestFixture(Config.Browser.iOS11point4)]
     [TestFixture(Config.Browser.iOS12)]
     [TestFixture(Config.Browser.iOS11point3)]
     [TestFixture(Config.Browser.Android6Marshmello)]
-    [TestFixture(Config.Browser.Android8point1Oreo)]
     [TestFixture(Config.Browser.Android7point1Nougat)]
     [TestFixture(Config.Browser.Android5point1Lolliopop)]
     class AutoCompletePageTest : BaseTest
--- a/test/UniformDocs.Tests/Test/FileUploadPageTest.cs
+++ b/test/UniformDocs.Tests/Test/FileUploadPageTest.cs
@@ -6,8 +6,19 @@ using UniformDocs.Tests.Utilities;
 namespace UniformDocs.Tests.Test
 {
     [TestFixture(Config.Browser.Chrome)]
-    [TestFixture(Config.Browser.Edge)] //BUG on EDGE https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7194303/
+    [TestFixture(Config.Browser.Edge)]
     [TestFixture(Config.Browser.Firefox)]
+    [TestFixture(Config.Browser.iOS11point2)]
+    [TestFixture(Config.Browser.Android7Nougat)]
+    [TestFixture(Config.Browser.Android8Oreo)]
+    [TestFixture(Config.Browser.iOS11point4)]
+    [TestFixture(Config.Browser.iOS12)]
+    [TestFixture(Config.Browser.iOS11point3)]
+    [TestFixture(Config.Browser.Android7Nougat)]
+    [TestFixture(Config.Browser.Android6Marshmello)]
+    [TestFixture(Config.Browser.Android8point1Oreo)]
+    [TestFixture(Config.Browser.Android7point1Nougat)]
+    [TestFixture(Config.Browser.Android5point1Lolliopop)]
     class FileUploadPageTest : BaseTest
     {
         private FileUploadPage _fileUploadPage;
--- a/test/UniformDocs.Tests/Test/FileUploadPageTest.cs
+++ b/test/UniformDocs.Tests/Test/FileUploadPageTest.cs
@@ -6,7 +6,7 @@ using UniformDocs.Tests.Utilities;
 namespace UniformDocs.Tests.Test
 {
     [TestFixture(Config.Browser.Chrome)]
-    [TestFixture(Config.Browser.Edge)]
+    [TestFixture(Config.Browser.Edge)] //BUG on EDGE https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7194303/
     [TestFixture(Config.Browser.Firefox)]
     class FileUploadPageTest : BaseTest
     {
--- a/test/UniformDocs.Tests/Test/FileUploadPageTest.cs
+++ b/test/UniformDocs.Tests/Test/FileUploadPageTest.cs
@@ -8,17 +8,6 @@ namespace UniformDocs.Tests.Test
     [TestFixture(Config.Browser.Chrome)]
     [TestFixture(Config.Browser.Edge)]
     [TestFixture(Config.Browser.Firefox)]
-    [TestFixture(Config.Browser.iOS11point2)]
-    [TestFixture(Config.Browser.Android7Nougat)]
-    [TestFixture(Config.Browser.Android8Oreo)]
-    [TestFixture(Config.Browser.iOS11point4)]
-    [TestFixture(Config.Browser.iOS12)]
-    [TestFixture(Config.Browser.iOS11point3)]
-    [TestFixture(Config.Browser.Android7Nougat)]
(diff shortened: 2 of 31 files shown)


test/UniformDocs.Tests/Utilities/WebDriverManager.cs
```diff
@@ -57,6 +57,107 @@ public static RemoteWebDriver StartDriver(Config.Browser browser, TimeSpan timeo
...
+                        capability.SetCapability("device", "iPhone SE");
+                        capability.SetCapability("real_mobile", "true");
+                        capability.SetCapability("project", "UniformDocs");
+                        capability.SetCapability("name", "TeamCity");
+                        capability.SetCapability("browserstack.local", "true");
+                        capability.SetCapability("browserstack.debug", "true");
+                        capability.SetCapability("browserstack.networkProfile", "3g-umts-good");
+                        capability.SetCapability("browserstack.video", "false");
+                        capability.SetCapability("browserstack.user", "marcinwarpechows2");
+                        capability.SetCapability("browserstack.key", "QgzxxRnmuB5yBrpv1yqW");
```
warpech: Much of these capabilities are not DRY. In fact, they are overwritten after line 163.

test/UniformDocs.Tests/Utilities/WebDriverManager.cs
```diff
@@ -57,6 +57,107 @@ public static RemoteWebDriver StartDriver(Config.Browser browser, TimeSpan timeo
                         capability.SetCapability("browser_version", "62.0");
                         break;
                     }
+                case Config.Browser.iPhoneSE:
+                    {
+                        capability = new DesiredCapabilities();
+                        capability.SetCapability("os_version", "11.2");
```
warpech: I don't agree with the tactics of testing in the outdated mobile OSs. 

For desktop browsers, we test in the current versions of the OS and web browser. Why for the mobile OS we should do the opposite?

warpech:
Note about selecting which browsers to run:

For TeamCity, the script https://github.com/Starcounter/UniformDocs/blob/master/tools/Teamcity.ps1 requires the browser names to be passed as string from the TC build config. 

By default that remains Chrome, Firefox, as can be seen here: https://teamcity.starcounter.org/admin/editRunType.html?id=template:Apps_Blending_Template&runnerId=RUNNER_17&cameFromUrl=%2Fadmin%2FeditBuildRunners.html%3Fid%3Dtemplate%253AApps_Blending_Template%26init%3D1&cameFromTitle=

image

----

For local testing, NUnit assumes to run the test by default in these browsers: https://github.com/Starcounter/UniformDocs/blob/add-mobile-support-tests/test/UniformDocs.Tests/Test/BaseTest.cs#L38-L39

It can be overridden by passing an argument to `test.bat`, as explained in `CONTRIBUTING.md`.

test/UniformDocs.Tests/Utilities/WebDriverManager.cs
alshakero: It seems BS don't support any newer iOS for iPhone SE https://www.browserstack.com/automate/capabilities

warpech: Yes, but I mean that the whole data is outdated, regarding both the OS versions and devices. The customer asked us to propose the list of supported browsers, so let's try to play that in our advantage, not in our disadvantage. BrowserStack's 2016 smartphone survey is not the best point of referrence regarding smartphones used in Sweden in 2019. 

We should aim at doing our best to support devices that we can actually verify and that are the most likely to be used by the end users.

Currently, the most popular device in the target market for our current project (Sweden) is iPhone 7. The most popular Android set is Galaxy S7, which is not being tested here. [Source](https://deviceatlas.com/blog/most-popular-smartphones#sweden)

Currently, the most popular Android device on that market is , which is not being tested here. [Source](https://deviceatlas.com/blog/most-popular-smartphones#sweden)

The most popular Android version in Sweden is Android 8, which is not reflected here. [Source](http://gs.statcounter.com/android-version-market-share/mobile-tablet/sweden)

For iPad, we should test iPad (2018) with Safari and iOS 12.

alshakero:
I used statcounter to get statistics for Sweden and hardcoded the most common [iOS](http://gs.statcounter.com/os-version-market-share/ios/mobile-tablet/sweden) and [Android ](http://gs.statcounter.com/os-version-market-share/android/mobile-tablet/sweden)devices in UD tests.

test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
```diff
@@ -7,6 +7,17 @@ namespace UniformDocs.Tests.Test
     [TestFixture(Config.Browser.Chrome)]
     [TestFixture(Config.Browser.Edge)]
     [TestFixture(Config.Browser.Firefox)]
+    [TestFixture(Config.Browser.iOS11point2)]
+    [TestFixture(Config.Browser.Android7Nougat)]
+    [TestFixture(Config.Browser.Android8Oreo)]
+    [TestFixture(Config.Browser.iOS11point4)]
+    [TestFixture(Config.Browser.iOS12)]
+    [TestFixture(Config.Browser.iOS11point3)]
+    [TestFixture(Config.Browser.Android7Nougat)]
```
warpech: This is a duplicate of value on line 11, causing the tests to fail

alshakero:
```suggestion
```

alshakero:
All tests passed!

test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
```diff
@@ -7,6 +7,13 @@ namespace UniformDocs.Tests.Test
     [TestFixture(Config.Browser.Chrome)]
     [TestFixture(Config.Browser.Edge)]
     [TestFixture(Config.Browser.Firefox)]
+    [TestFixture(Config.Browser.iOS11point2)]
```
warpech: I'd rather use an underscore instead of the word "point", e.g. `iOS11_2`

test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
```diff
@@ -7,6 +7,13 @@ namespace UniformDocs.Tests.Test
...
+    [TestFixture(Config.Browser.Android7Nougat)]
```
warpech: I'd rather not duplicate information about the codename of the Android version. `Android7` is sufficient.

test/UniformDocs.Tests/Test/FileUploadPageTest.cs
```diff
@@ -6,7 +6,7 @@
 namespace UniformDocs.Tests.Test
 {
     [TestFixture(Config.Browser.Chrome)]
-    [TestFixture(Config.Browser.Edge)] //BUG on EDGE https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7194303/
+    [TestFixture(Config.Browser.Edge)]
```
warpech: What's the reason of removing this comment?

test/UniformDocs.Tests/Utilities/Config.cs
```diff
@@ -29,10 +36,9 @@ public enum Buttons
        f => f.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork)
    .ToString();
         public static readonly Uri TestedAppUrl = new Uri($"http://{LocalIP}:8080/UniformDocs");
-        public static readonly Uri RemoteWebDriverUri = new Uri("http://hub-cloud.browserstack.com/wd/hub/");
-        public static readonly string BrowserstackUsername = Environment.GetEnvironmentVariable("BROWSERSTACK_USERNAME");
-        public static readonly string BrowserstackAccessKey = Environment.GetEnvironmentVariable("BROWSERSTACK_ACCESS_KEY");
-        public static readonly string BrowserstackLocalIdentifier = Environment.GetEnvironmentVariable("COMPUTERNAME");
+        public static readonly string SauceLabsUserName = Environment.GetEnvironmentVariable("SAUCE_USERNAME");
+        public static readonly string SauceAccessKey = Environment.GetEnvironmentVariable("SAUCE_ACCESS_KEY");
```
warpech: Inconsistent with the name on line 39. Should be `SauceLabsAccessKey`

test/UniformDocs.Tests/Utilities/WebDriverManager.cs
```diff
@@ -51,29 +47,68 @@ public static RemoteWebDriver StartDriver(Config.Browser browser, TimeSpan timeo
...
-                capability.SetCapability("browserstack.user", Config.BrowserstackUsername);
-                capability.SetCapability("browserstack.key", Config.BrowserstackAccessKey);
-                capability.SetCapability("browserstack.debug", "true");
-                capability.SetCapability("browserstack.local", "true");
-                capability.SetCapability("browserstack.localIdentifier", Config.BrowserstackLocalIdentifier);
-                capability.SetCapability("browserstack.video", "false"); //video recording increases test time, see https://www.browserstack.com/automate/c-sharp
-                capability.SetCapability("browserstack.use_w3c", "true"); //needed for closing of window in UrlPage_ClickBlankTargettedLink for Firefox on BrowserStack; see https://github.com/SeleniumHQ/selenium/issues/5064
-                capability.SetCapability("browserstack.console", "verbose");
+                capability.SetCapability("recordVideo", false);
+                capability.SetCapability("recordScreenshots", false);
```
warpech: That's a regression from BrowserStack, where we do record screenshots. Is there any reason for it?

test/UniformDocs.Tests/Utilities/WebDriverManager.cs
```diff
@@ -51,29 +47,68 @@ public static RemoteWebDriver StartDriver(Config.Browser browser, TimeSpan timeo
...
             }
 
             capability.SetCapability("project", "UniformDocs");
```
warpech: `project` is not supported by SauceLabs. Maybe use `capability.SetCapability("tags", new List<string> { "UniformDocs" });`?

test/UniformDocs.Tests/Utilities/WebDriverManager.cs
```diff
@@ -51,29 +47,68 @@ public static RemoteWebDriver StartDriver(Config.Browser browser, TimeSpan timeo
...
             {
                 IsCloud = true;
-                capability.SetCapability("browserstack.user", Config.BrowserstackUsername);
-                capability.SetCapability("browserstack.key", Config.BrowserstackAccessKey);
-                capability.SetCapability("browserstack.debug", "true");
-                capability.SetCapability("browserstack.local", "true");
-                capability.SetCapability("browserstack.localIdentifier", Config.BrowserstackLocalIdentifier);
-                capability.SetCapability("browserstack.video", "false"); //video recording increases test time, see https://www.browserstack.com/automate/c-sharp
-                capability.SetCapability("browserstack.use_w3c", "true"); //needed for closing of window in UrlPage_ClickBlankTargettedLink for Firefox on BrowserStack; see https://github.com/SeleniumHQ/selenium/issues/5064
-                capability.SetCapability("browserstack.console", "verbose");
```
warpech: That's a regression from BrowserStack, because you did not set up SauceLabs to record console logs. Use `capability.SetCapability("extendedDebugging", "true");`

test/UniformDocs.Tests/Utilities/WebDriverManager.cs
```diff
@@ -97,37 +132,8 @@ public static void MarkTestStatusOnBrowserStack(RemoteWebDriver driver, ResultSt
...
-
-            NetworkCredential myNetworkCredential = new NetworkCredential(Config.BrowserstackUsername, Config.BrowserstackAccessKey);
-            CredentialCache myCredentialCache = new CredentialCache();
-            myCredentialCache.Add(myUri, "Basic", myNetworkCredential);
-            myHttpWebRequest.PreAuthenticate = true;
-            myHttpWebRequest.Credentials = myCredentialCache;
-
-            myWebRequest.GetResponse().Close();
+                       
+            ((IJavaScriptExecutor)driver).ExecuteScript("sauce:job-result=" + (outcome.Equals(ResultState.Success) ? "passed" : "failed"));
```
warpech: The method should be change to not have the name `BrowserStack` in it.

test/UniformDocs.Tests/Utilities/Config.cs
```diff
@@ -29,10 +36,9 @@ public enum Buttons
...
+        public static readonly Uri RemoteWebDriverUri = new Uri($"http://{SauceLabsUserName}:{SauceAccessKey}@ondemand.saucelabs.com:80/wd/hub");
```
warpech: It's a better practice to send credentials in capabilties, `username` and `accessKey`.

warpech:
I pushed a commit that addresses some of my comments. Pls check if you agree.

Some additional comments:

1. Tests seem to execute longer in SauceLabs than in BrowserStack. What will be de the total time of Chrome + Firefox tests? What will be the total time of all tests using the current config?
2. Many tests fail under certain mobile browsers, for example `UniformDocs.Tests.Test.AutoCompletePageTest(Android7point1Nougat).AutoCompletePage_FillProductNameThenSelectProduct`. We must not merge a PR that has failing tests.
3. How do you find the UI of SauceLabs? For me it looks inferior to BrowserStack, because page events, console logs, etc, are not shown on a single timeline. Or maybe I am using it wrong? I am afraid debugging on SauceLabs will be much poorer. What's your opinion on this?
4. Instructions in CONTRIBUTING.md need to be updated for SauceLabs

test/UniformDocs.Tests/Test/FileUploadPageTest.cs
alshakero: Probably a blunder

alshakero:
> Tests seem to execute longer in SauceLabs than in BrowserStack. What will be de the total time of Chrome + Firefox tests? What will be the total time of all tests using the current config?

I don't really know. I think we'll have to measure it. But something fundementally different happened when we migrated to Browserstack and is still happening with Saucelabs; every test unit is launching a browser instance and testing it then it's killing it and the whole VM. Every single unit is starting a VM + a browser. Whereas when we tested locally, Chrome would start and streak through a dozen tests the closes. waaay faster. Any idea if we could go back to that but on a remote testing env? I almost feel this should be a show stopper in comparision to before. It took us 12min when we tested locally. Now it's taking 53mins (on BrowserStack).

> Many tests fail under certain mobile browsers, for example UniformDocs.Tests.Test.AutoCompletePageTest(Android7point1Nougat).AutoCompletePage_FillProductNameThenSelectProduct. We must not merge a PR that has failing tests.

Yes sure I agree, but I didn't want to work hard on this before SauceLabs idea is approved. 

> How do you find the UI of SauceLabs? For me it looks inferior to BrowserStack, because page events, console logs, etc, are not shown on a single timeline. Or maybe I am using it wrong? I am afraid debugging on SauceLabs will be much poorer. What's your opinion on this?

I find them both OK. BrowserStack is certainly better but I'm very well accustomed to SauceLabs and know how to move around. 

> Instructions in CONTRIBUTING.md need to be updated for SauceLabs
I'll do that once we agree for SauceLabs.

warpech:
> I don't really know. I think we'll have to measure it. But something fundementally different happened when we migrated to Browserstack and is still happening with Saucelabs; every test unit is launching a browser instance and testing it then it's killing it and the whole VM. Every single unit is starting a VM + a browser. Whereas when we tested locally, Chrome would start and streak through a dozen tests the closes. waaay faster. Any idea if we could go back to that but on a remote testing env? I almost feel this should be a show stopper in comparision to before. It took us 12min when we tested locally. Now it's taking 53mins (on BrowserStack).

This is not correct. Every test **class fixture* is a separate VM + a browser. Tests within a class are run within the same connection. That's why you see that the first test in a class tests longer than the other tests. I don't have an idea how to speed it up, other than do more work in parallel (https://github.com/Starcounter/UniformDocs/issues/387)

alshakero:
Closing as obselete

Issue #394, opened 2019-03-01
[Urgent] Remove the use of ScheduleTask


warpech:
This app fails to build with Starcounter 2.4.0.7177 and 2.4.990.239 because it uses `ScheduleTask` method, which was marked as "Obsolete" since Nov 2017 and removed in Feb 2019.

The change in Starcounter that removes `ScheduleTask`: https://github.com/Starcounter/level1/pull/4891

-----

### Steps to reproduce the problem in UniformDocs

1. Start UniformDocs
2. Go to the page http://localhost:8080/UniformDocs/ProgressBar
3. Click on the button "Download (imaginary) file"

### Expected

The progress bar should perform an animation to 100%

### Actual

There is an exception that can be seen in the browser console and Administrator Log:

```
System.MissingMethodException: Method not found: 'Void Starcounter.Session.ScheduleTask(System.String, SessionTask, Boolean)'.
   at Joozek78.Star.Async.Internal.StarcounterSessionSynchronizationContext.Post(SendOrPostCallback d, Object state)
   at System.Threading.Tasks.AwaitTaskContinuation.RunCallback(ContextCallback callback, Object state, Task& currentTask)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
HResult=-2146233069
```

Starcounter version 2.4.990.239  
UniformDocs version `702ae87f89d1b6499ed7e1adac0bfb188cf3540c`

warpech:
Is there any benefit of using `Joozek78.Star.Async`? We need to either fork and fix this library or switch to use `Scheduling.RunTask` directly.

kegor:
Related bug https://github.com/Starcounter/UniformDocs/issues/300.

kegor:
I asked @iqbalskd to help us with the configuring CI for the possibility to release patched package.

kegor:
Closing because all related PRs were merged.

PR #397, opened 2019-03-04
Problem 394: temporary fix for removing using ScheduleTask.

kegor:
For #394.

Were tested `build&run` and the general case on the `ProgressBar` page.


kegor:
@PatrykSzwer checked `build&run` and then `ProgressBar` page, it works ok (but after `msbuild /t:clean`), thank you for double checking.

Issue #395, opened 2019-03-01
Improve test error reporting when the app crashes


warpech:
When the app crashes, currently NUnit continues to run the Selenium tests, all of which report that an element cannot be found. This is very misleading for the person investigating the failure.

Instead, I would like NUnit to quit testing with the message "The tested app [AppName] is not running", ideally with the last error message from the Starcounter Error Log.

The message can be obtained using the CLI command listed at https://docs.starcounter.io/topic-guides/working-with-starcounter/error-log or via a REST API call: 

```
curl -sS "http://localhost:8181/api/admin/log?debug=false&error=true&maxitems=3&notice=false&source=&warning=true" | python -c 'import json,sys;print(json.load(sys.stdin)["LogEntries"][1]["Message"])'
```

Though I am not sure that there will always be a message and that the message will always be relevant.

warpech:
Reassigning this issue from Egor to me. I will see if I can finish the https://github.com/Starcounter/UniformDocs/pull/398

Issue #75, opened 2016-11-07
Flaky tests in KitchenSink


joozek78:
Flaky test: TextPageTest("firefox").TextPage_TextPropagationOnUnfocus
Go to https://teamcity.starcounter.org/project.html?projectId=Starcounter_Apps_KitchenSink&testNameId=7318267521195939949&tab=testDetails
 and see that this test sometimes fails for no apparent reason. Check if you can find the reason for this (race condition?) and fix if possible. If you can't find the reason after investigation please close this issue with a proper comment

joozek78:
Another flaky test: ValidationPageTest("firefox").ValidationPage_ValidRequireInput

warpech:
Fixed with PR #83, thanks for reporting

PR #74, opened 2016-11-01
Pagination issue #72

Mackiovello:
#72

PR #398, opened 2019-03-08
Problem 395: Improve test error reporting when the app crashes.

kegor:
**Problem #395:**
>My proposal is like this:
>
>1. Before each test, check if the app is running. If not, quit with an error that the app is not running
>2. After each test, check if the app is running. If not, quit with an error that the app crashed with the last error messages from the log.

**Solution:**
1. Was added helper class for getting needed kind-of `administration` info through the REST API.
2. For the first iteration (to discuss current implementation) were added invocations of the `Assert.Inconclusive` method with required message inside.

**Tests:**
1. `D:\git\UniformDocs>test --test="UniformDocs.Tests.Test.ProgressBarTest"` without any other actions - passed.
2. `D:\git\UniformDocs>test --test="UniformDocs.Tests.Test.ProgressBarTest"` and stopping the `UniformDocs` app after 10 seconds - didn't passed.

**Screenshots:**
tests4
tests1

**Questions:**
1. @warpech maybe we should use `--stoponerror` command instead of throwing the exception in the current test using `Assert.Inconclusive`? Or from other hand this will break the module structure pattern  of the unit-tests?
2. Maybe you have an idea about another test case? (i already got almost similar result starting tests in VS)
3. For now i don't have the solid answer why the **first error** (on the first screenshot) talking about `null reference exception` in the `[OneTimeTearDown]` method and how bad this for us.

### DIFF
--- a/test/UniformDocs.Tests/Test/BaseTest.cs
+++ b/test/UniformDocs.Tests/Test/BaseTest.cs
@@ -20,6 +20,7 @@ namespace UniformDocs.Tests.Test
         private List<string> _browsersToRun = new List<string>();
         private ResultState LastOutcome;
         private string LastOutcomeMessage;
+        private static readonly string appName = "UniformDocs";
 
         public BaseTest(Config.Browser browser)
         {
@@ -60,6 +61,11 @@ namespace UniformDocs.Tests.Test
         [TearDown]
         public void TearDown()
         {
+            if (!(RestApiHelper.CheckAppRunning(appName).Result))
+            {
+                Assert.Inconclusive($"The tested app {appName} crushed. The possible reason: {RestApiHelper.GetLatestLogEntry().Result}");
+            }
+
             if (LastOutcome == null || TestContext.CurrentContext.Result.Outcome != ResultState.Success)
             {
                 //this class has single driver session for all test methods
--- a/test/UniformDocs.Tests/Test/ProgressBarTest.cs
+++ b/test/UniformDocs.Tests/Test/ProgressBarTest.cs
@@ -12,6 +12,7 @@ namespace UniformDocs.Tests.Test
     {
         private ProgressBarPage _progressBarPage;
         private MainPage _mainPage;
+        private static readonly string appName = "UniformDocs";
 
         public ProgressBarTest(Config.Browser browser) : base(browser)
         {
@@ -20,6 +21,11 @@ namespace UniformDocs.Tests.Test
         [SetUp]
         public void SetUp()
         {
+            if (!(RestApiHelper.CheckAppRunning(appName).Result))
+            {
+                Assert.Inconclusive($"The tested app {appName} is not running");
+            }
+
             _mainPage = new MainPage(Driver).GoToMainPage();
             _progressBarPage = _mainPage.GoToProgressBarPage();
         }
--- a/test/UniformDocs.Tests/Utilities/RestApiHelper.cs
+++ b/test/UniformDocs.Tests/Utilities/RestApiHelper.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Starcounter.Internal;
+
+namespace UniformDocs.Tests.Utilities
+{
+    public static class RestApiHelper
+    {
+        private static readonly ushort InternalPort = StarcounterEnvironment.Default.SystemHttpPort;
+        private const string InternalHost = "127.0.0.1";
+
+        public class DatabaseApplicationsJson
+        {
+            public IEnumerable<App> Items { get; set; }
+        }
+
+        public class Log
+        {
+            public IEnumerable<LogEntry> LogEntries { get; set; }
+        }
+
+        public class App
+        {
+            public string DisplayName { get; set; }
+        }
+
+        public class LogEntry
+        {
+            public string Message { get; set; }
+        }
+
+        public static async Task<bool> CheckAppRunning(string appName)
+        {
+            HttpResponseMessage response = await new HttpClient().GetAsync($"http://{InternalHost}:{InternalPort}/api/admin/databases/default/applications");
+
+            if (response.IsSuccessStatusCode)
+            {
+                var runningAppsString = await response.Content.ReadAsStringAsync();
+                var runningApps = JsonConvert.DeserializeObject<DatabaseApplicationsJson>(runningAppsString);
+
+                foreach (var app in runningApps.Items)
+                {
+                    if (app.DisplayName == appName)
+                    {
+                        return true;
+                    }
+                }
+            }
+
+            return false;
+        }
+
+        public static async Task<string> GetLatestLogEntry()
+        {
+            string Parameters = "debug=false&error=true&maxitems=1&notice=false&source=&warning=true";
+
+            HttpResponseMessage response = await new HttpClient().GetAsync($"http://{InternalHost}:{InternalPort}/api/admin/log?{Parameters}");
+
+            if (response.IsSuccessStatusCode)
+            {
+                var latestLogEntryString = await response.Content.ReadAsStringAsync();
+
+                var latestLogEntry = JsonConvert.DeserializeObject<Log>(latestLogEntryString);
+
+                if (latestLogEntry.LogEntries.FirstOrDefault() != null && !string.IsNullOrEmpty(latestLogEntry.LogEntries.First().Message))
+                {
+                    return latestLogEntry.LogEntries.First().Message;
+                }
+            }
+
+            return String.Empty;
+        }
+    }
+}
--- a/test/UniformDocs.Tests/UniformDocs.Tests.csproj
+++ b/test/UniformDocs.Tests/UniformDocs.Tests.csproj
@@ -49,6 +49,7 @@
     <Reference Include="Starcounter.Internal, Version=2.0.0.0, Culture=neutral, PublicKeyToken=d2df1e81d0ca3abf, processorArchitecture=MSIL" />
     <Reference Include="System" />
     <Reference Include="System.Drawing" />
+    <Reference Include="System.Net.Http" />
     <Reference Include="WebDriver, Version=3.11.2.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\..\packages\Selenium.WebDriver.3.11.2\lib\net45\WebDriver.dll</HintPath>
     </Reference>
@@ -113,6 +114,7 @@
     <Compile Include="Ui\TextareaPage.cs" />
     <Compile Include="Ui\TextPage.cs" />
     <Compile Include="Utilities\Config.cs" />
+    <Compile Include="Utilities\RestApiHelper.cs" />
     <Compile Include="Utilities\WebDriverManager.cs" />
   </ItemGroup>
   <ItemGroup>
--- a/test/UniformDocs.Tests/packages.config
+++ b/test/UniformDocs.Tests/packages.config
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="Microsoft.Net.Http" version="2.2.29" targetFramework="net461" />
   <package id="Newtonsoft.Json" version="8.0.3" targetFramework="net461" />
   <package id="NUnit" version="3.10.1" targetFramework="net461" />
   <package id="NUnit.ConsoleRunner" version="3.6.1" targetFramework="net461" />


test/UniformDocs.Tests/Test/BaseTest.cs
```diff
@@ -20,6 +20,7 @@ public class BaseTest
         private List<string> _browsersToRun = new List<string>();
         private ResultState LastOutcome;
         private string LastOutcomeMessage;
+        private static readonly string appName = "UniformDocs";
```
PatrykSzwer: Maybe you could replace this with just getting the name from the app itself? I think there is a way to do that. You can check this e.g: https://stackoverflow.com/questions/14829689/how-to-get-exe-application-name-and-version-in-c-sharp-compact-framework/14843617.

test/UniformDocs.Tests/Test/ProgressBarTest.cs
```diff
@@ -20,6 +21,11 @@ public ProgressBarTest(Config.Browser browser) : base(browser)
         [SetUp]
         public void SetUp()
         {
+            if (!(RestApiHelper.CheckAppRunning(appName).Result))
+            {
+                Assert.Inconclusive($"The tested app {appName} is not running");
```
PatrykSzwer: Again, `Assert.Inconclusive` is not the right thing to use here. I'd rather go with either `Assert.Fail` or throwing an exception.

test/UniformDocs.Tests/Test/ProgressBarTest.cs
```diff
@@ -12,6 +12,7 @@ class ProgressBarTest : BaseTest
     {
         private ProgressBarPage _progressBarPage;
         private MainPage _mainPage;
+        private static readonly string appName = "UniformDocs";
```
warpech: This is redundant, because the same field was added in `BaseTest.cs`

test/UniformDocs.Tests/Test/BaseTest.cs
warpech: This will get `UniformDocs.Tests` as the app name, so unfortunately it can be used.

For clarity, I propose to change the field name from `appName` to `_testedAppName`.

test/UniformDocs.Tests/Test/BaseTest.cs
```diff
@@ -60,6 +61,11 @@ public void TestFixtureSetUp()
         [TearDown]
         public void TearDown()
         {
+            if (!(RestApiHelper.CheckAppRunning(appName).Result))
+            {
+                Assert.Inconclusive($"The tested app {appName} crushed. The possible reason: {RestApiHelper.GetLatestLogEntry().Result}");
```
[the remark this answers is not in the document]
warpech: 1. According to https://stackoverflow.com/questions/1219274/usage-of-assert-inconclusive, this looks like an adequate use of `Inconclusive`. I believe that it goes like this:
   - `Fail` - the test failed because the output it different than expected
   - `Ignore` - the test should not run because of a decision made before launching the tests
   - `Inconclusive` - the test did not run because an external condition was not met
2. Typo: `crushed` -> `crashed`.
3. @kegor wrote "possible" reason, because he links the crash of the app with the most recent entry from the error log. Which is not guaranteed to be correct if the app crashed without an error. 
4. For clarity, I propose to change the message to:

```
The tested app {appName} unexpectedly stopped during the test. The 
following text is the last message that can be found in the Starcounter log: 
{RestApiHelper.GetLatestLogEntry().Result}
```

test/UniformDocs.Tests/Test/ProgressBarTest.cs
warpech: I think `Inconclusive` is OK as long as it has the same consequences as `Fail`.

test/UniformDocs.Tests/Utilities/RestApiHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Starcounter.Internal;
+
+namespace UniformDocs.Tests.Utilities
+{
+    public static class RestApiHelper
+    {
+        private static readonly ushort InternalPort = StarcounterEnvironment.Default.SystemHttpPort;
+        private const string InternalHost = "127.0.0.1";
```
warpech: I think the two above properties should be moved to `Config.cs`.

warpech:
Can you pls provide a screenshot how is an inconclusive test reported in TeamCity and BrowserStack?

Does having a single inconclusive test mark the whole build as failing in TC (it should)?

> 1. @warpech maybe we should use `--stoponerror` command instead of throwing the exception in the current test using `Assert.Inconclusive`? Or from other hand this will break the module structure pattern  of the unit-tests?

I think we can use both `Assert.Inconclusive` and `--stoponerror`, no? This means the tests will break after first `Inconclusive` or first `Fail`, right? @PatrykSzwer, @kegor WDYT about that? I think this would be cool for TC. For local testing, someone can choose to not this command if they don't like it.

> 2. Maybe you have an idea about another test case? (i already got almost similar result starting tests in VS)

I don't understand, another test case for what? You mean what else should we check apart from is the SUT running?

> 3. For now i don't have the solid answer why the **first error** (on the first screenshot) talking about `null reference exception` in the `[OneTimeTearDown]` method and how bad this for us.

I guess this means if the test is indecisive, the `outcome` parameter to `MarkTestStatusOnBrowserStack` is `null`, in which case we should report to BrowserStack accordingly. Sadly, BrowserStack understands only two possible states: https://www.browserstack.com/automate/rest-api

test/UniformDocs.Tests/Test/BaseTest.cs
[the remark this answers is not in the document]
PatrykSzwer: >The
following text is the last message that can be found in the Starcounter log:
{RestApiHelper.GetLatestLogEntry().Result}

Thaaats the solution I was hoping for 👍

PatrykSzwer:
>I think we can use both Assert.Inconclusive and --stoponerror, no? **This means the tests will break after first Inconclusive or first Fail**, right? @PatrykSzwer, @kegor WDYT about that?

I like that idea, let's do it!

kegor:
1. Sounds logical, on TC we actually only interested to check that build is green and mention only one (first) failed test is enough for investigation, and locally we could be interested to run all tests the check them all in the background. So i agree, my first idea regarding implementation is to run UniformDocs with special argument if we would like to run all tests (disregarding first and consequential fails), and by default we can use `--stoponerror`.
2.  I meant another ways for manual tests.
3.  So in this case (`LastOutcome == null`) it will be better to send `failed` state?

4/5. And now I'm thinking how to reproduce situation with crashed app on TC and how to what kind of BrowserStack report i could check, @warpech maybe you could bump me with hints?

PatrykSzwer:
>4/5. And now I'm thinking how to reproduce situation with crashed app on TC and how to what kind of BrowserStack report i could check, @warpech maybe you could bump me with hints?

Just create some tmp branch and push a commit with a test that (on purpose) is resulting in inconclusive state (just return `Assert.Inconclusive` in a test). Then, run a build of this branch on teamcity and look at the results.

kegor:
Results of the tests with `Assert.Inconclusive` and `--stoponerror`:
1. Using `Assert.Inconclusive` in the start of the second of the `UniformDocs.Tests.Test.ProgressBarTest` on TC:
image
And overall result of the build is - ✅ (https://teamcity.starcounter.org/viewLog.html?buildId=93039&tab=queuedBuildOverviewTab).
2. Using  `Assert.Inconclusive` in the start of the first of the `UniformDocs.Tests.Test.ProgressBarTest` locally with `--stoponerror` argument for the `nunit3-console`:
image
3. Using  `throw new Exception()` in the start of the first of the `UniformDocs.Tests.Test.ProgressBarTest` locally with `--stoponerror` argument for the `nunit3-console`:
image

As the interim opinion i would add `--stoponerror` on https://github.com/Starcounter/UniformDocs/blob/master-2.4/tools/Teamcity.ps1#L48 and use regular throwing exceptions instead of `Assert.Inconclusive`. WDYT @warpech, @PatrykSzwer? Did i missed any cases to check additionally?

warpech:
> 3. So in this case (LastOutcome == null) it will be better to send failed state?

Yeah, definitely `failed` in this case.

----

I think this sucks that NUnit (or is it TC?) treats `Inconclusive` as passing.

Then I guess we need to fail the tests if the app is not running. Isn't `throw new Exception()` equivalent to `Assert.Fail()`, anyway?

PatrykSzwer:
Looking at this build, is really hard to decide if with the build is something wrong or not. The only clue is the `Ignored` status of those tests: 
image

I agree that this is definitely no-go, and those tests should be marked as failed ones by teamcity. 
>Isn't throw new Exception() equivalent to Assert.Fail(), anyway?

Not exactly. `Assert.Fail()` throws `AssertFailedException`, so pretty specific one.

kegor:
@warpech, @PatrykSzwer 
3. i've changed the line https://github.com/Starcounter/UniformDocs/blob/master/test/UniformDocs.Tests/Utilities/WebDriverManager.cs#L116 like `if (outcome != null && outcome.Equals(ResultState.Success))` (for the moment locally, because probably i will make more changes and the push might be more saturated)

----
I tried to use `Assert.Fail()` instead of `throw new Exception()` at the start of `UniformDocs.Tests.Test.ProgressBarTest.ProgressBarPage_ProgressBar_DownloadFile` locally with `--stoponerror` argument for the nunit3-console and got the same result as on the screenshot above (with failed (❌) result).

Then i quickly tried to find the answer why using `Assert.Inconclusive` with `--stoponerror` doesn't leads to the same result despite the fact that `Assert.Inconclusive` also throws an exception (https://docs.microsoft.com/en-us/dotnet/api/nunit.framework.assert.inconclusive) but i didn't, probably the reason is in the type of this exception https://docs.microsoft.com/en-us/dotnet/api/nunit.framework.inconclusiveexception.

warpech:
1. Agreed
2. @PatrykSzwer had an argument that this might be unwanted. I agree with him now, sometimes there is a change that causes, say, 3 tests to fail. Having the test framework to stop at the 1st error might become difficult to work with. 

I wish they had something like this: https://github.com/nunit/nunit-console/issues/169

warpech:
@kegor what's the status of this?

kegor:
@warpech, I remembered that i decided to wait a bit `NUnit`-team response here https://github.com/nunit/nunit-console/issues/169, but now i think we can go further without using `--stoponerror` while we can't specify the number of errors before tests stop and only add following change `Use Assert.Fail instead of Assert.Inconclusive.` to the already locally existed `if (outcome != null && outcome.Equals(ResultState.Success))`?

warpech:
This allows to stop NUnit for example after 10 failing tests? If yes, then I am all for it. Great idea!

warpech:
To clarify my above comment, I think that if you can read the result of test in the `[Teardown]` method, the number of failures can be stored in a static variable. Then, after the limit is reached, break further test execution. This can be done by throwing an exception in the `[Setup]` method, if there is no better way. This makes `--stoponerror` not needed

kegor:
@warpech, @PatrykSzwer let me double check regarding `Assert.Inconclusive()` vs `Assert.Fail()`, how it looks in VS firstly:

1. `Assert.Inconclusive()`:

tests_stopped_as_inconclusive

_Pros:_
- tests were not actually performed, so it looks more logical;
_Cons:_
- the test class itself is marked as successful although the tests were not performed;

2. `Assert.Fail()`:

tests_stopped_as_failed

_Pros:_
- unstarted tests should be more noticeable in the TC (related to this https://github.com/Starcounter/UniformDocs/pull/398#issuecomment-471947968 and will be checked twice)
_Cons:_
- tests were not actually performed, so marking them as failed looks not so logical;

What do you think about this and for other changes (while i'm waiting the TC results).

warpech:
I slightly prefer `Assert.Inconclusive`, but I could accept it either way. The whole build will be marked as failed, because there are enough errors that caused the maximum number of errors to be reached.

kegor:
I made few more googling attempts and few more local tests in VS:

1. If some tests are inconclusive the whole test class is green:
some_tests_are_inconclusive

2. If all tests were marked as inconclusive in `[Setup]` method then whole test class is still green:
all_tests_are_inconclusive_setup

3. And if we calling `Assert.Inconclusive()` from `[OneTimeSetup]` method (i simply replaced our `Assert.Ignore()` method for the `Edge` to `Assert.Inconclusive()` for testing purpose) then test class is failed (red):
all_tests_are_inconclusive_onetimesetup

Looking at this question https://stackoverflow.com/questions/4733592/assert-inconclusive-and-ignoreattribute i started thinking that probably the color and overall result for the test class is the VS/Resharper/TeamCity specific thing, so probably i should perform the same tests on TeamCity because we mostly would like to simplify our investigations there?

warpech:
Yes, I can see that this greatly depends on the tooling. In my VS2017, any test class with at least one `Assert.Inconclusive` is marked as with a blue "information" icon, which is neither passed nor failed. It doesn't matter if I put `Assert.Inconclusive` in a test, setup, or one-time-setup.

I think it is OK to proceed with using `Assert.Inconclusive` when the maximum failure limit is reached. 

As I said, TeamCity will fail not because of how it treats inconclusive tests, but simply because there is more than 0 failures.

test/UniformDocs.Tests/Test/BaseTest.cs
```diff
@@ -20,6 +20,8 @@ public class BaseTest
         private List<string> _browsersToRun = new List<string>();
         private ResultState LastOutcome;
         private string LastOutcomeMessage;
+        protected static readonly string _testedAppName = "UniformDocs";
```
warpech: Can this be moved to `Config.cs`?

test/UniformDocs.Tests/Test/BaseTest.cs
```diff
@@ -20,6 +20,8 @@ public class BaseTest
...
+        private static int failsCount = 0;
```
warpech: I think it should be `_failsCount`

test/UniformDocs.Tests/Test/BaseTest.cs
```diff
@@ -57,9 +59,35 @@ public void TestFixtureSetUp()
             }
         }
 
+        [SetUp]
+        public virtual void SetUp()
+        {
+            if (failsCount >= Config.FailsBeforeStop)
+            {
+                Assert.Inconclusive("The maximum number of test errors was reached. Tests will be stopped.");
```
warpech: Maybe give the number of fails within the message and say that we will mark as inconclusive, i.e. `The maximum number of {failsCount} test errors was reached. The further tests are marked as inconclusive.`?

warpech:
Sometimes `failsCount` is not incremented.

### Steps to reproduce

1. Start BrowserStackLocal using the instructions in CONTRIBUTING.md
2. Start UniformDocs
3. Launch the tests
4. Stop the database

### Expected

After `failsCount` tests, all other tests should be inconclusive

### Actual

`failsCount` remain at zero, because it is not increased when `Assert.Fail($"The tested app {_testedAppName} is not running");` or `XXXX` happens.

test/UniformDocs.Tests/Test/BaseTest.cs
```diff
@@ -30,6 +30,11 @@ public BaseTest(Config.Browser browser)
         [OneTimeSetUp]
         public void TestFixtureSetUp()
         {
+            if (_failsCount >= Config.FailsBeforeStop)
```
kegor: Without this same (as in `[SetUp]` method) check here i see that every child tests actually is called in BrowswerStack and waits for 90 sec to be marked as timed out.

warpech:
Trying to reproduce the problem explained in https://github.com/Starcounter/UniformDocs/pull/398#issuecomment-485992403 still prevails, because of an unhandled exception in the method `RestApiHelper.CheckAppRunning`:

```
System.AggregateException
  HResult=0x80131500
  Message=One or more errors occurred.
  Source=mscorlib
  StackTrace:
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()
   at UniformDocs.Tests.Test.BaseTest.TearDown() in W:\repo\UniformDocs\test\UniformDocs.Tests\Test\BaseTest.cs:line 84

Inner Exception 1:
HttpRequestException: An error occurred while sending the request.

Inner Exception 2:
WebException: Unable to connect to the remote server

Inner Exception 3:
SocketException: No connection could be made because the target machine actively refused it
```

image

kegor:
@warpech, thank you for your investigation, i'll try to handle such case properly!

test/UniformDocs.Tests/Utilities/RestApiHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+                            return false;
+                        }
+                    }
+                    catch (Exception)
+                    {
+                        numberOfAttempts++;
+
+                        if (numberOfAttempts >= MaxRetries)
+                        {
+                            throw;
```
kegor: For now we re-throwing the full stack of the inner exceptions, but alternatively we could throw general exception with meaningful message.

Also i was thinking about the `Task.Delay` between the loops but i didn't find such approach in the related best practices, maybe because if we already have this problem then it is most likely won't be solved on the other side by the seconds.

warpech:
I will test the solution with 4 different scenarios:

### Steps to reproduce

1. Start BrowserStackLocal using the instructions in CONTRIBUTING.md
2. Start UniformDocs
3. Launch the tests
4. Try multiple connection failure scenarios:
   1. Normal assertion errors in actual tests 
   2. Stop the app `staradmin stop app UniformDocs` ✔️
   3. Stop the db `staradmin stop db default` ✔️
   4. Stop the server `staradmin kill devall` ❌ (see https://github.com/Starcounter/UniformDocs/pull/398#discussion_r279693892)
   5. Stop BrowserStackLocal (CTRL+C) ✔️

test/UniformDocs.Tests/Utilities/RestApiHelper.cs
kegor: @warpech this is my thoughts regarding the idea with wrapping all methods in `try/catch`:
1. For me it is not reasonable enough to wrap all code even that can't throw any exceptions.
2. On my view we going across to the logic that we have only few places that should leads to the stopping tests, and one of them is the fallen app.

This why i sent the iteration in which we increment our counter if we catching exception during the HttpClient call.

What i got for the case : >4. Stop the server `staradmin kill devall`:
image

I also checked other cases.

test/UniformDocs.Tests/Test/BaseTest.cs
```diff
@@ -81,7 +81,7 @@ public virtual void SetUp()
         [TearDown]
         public void TearDown()
         {
-            if (!(RestApiHelper.CheckAppRunning(Config.TestedAppName).Result))
```
warpech: I am not arguing, but don't you think that wrapping the body of `[TearDown]` and `[OneTimeTearDown]` in a try/catch would be a more elegant solution?

warpech:
I am not sure if this is something we can fix, but when I call scenario 4.iii (`staradmin stop db default`) I get slightly different results than expected.

See: https://teamcity.starcounter.org/viewLog.html?buildTypeId=Bifrost_Webplatform_UniformDocs_IntegrationUniformDocsDebug&buildId=96240&tab=testsInfo&pager.currentPage=10

#### Expected

I am expecting to see 5 errors, and specifically: The first test should fail with `The tested app UniformDocs unexpectedly stopped during the test.`, then 4 tests should fail with `The tested app UniformDocs is not running`, finally all the other tests should be ignored.

#### Actual

There are only 4 failing tests, and the messages are different than expected:

- First failing test (#73 in the Order column) fails with `Multiple failures or warnings in test`
- Next test in order (#74) fails with `Multiple failures or warnings in test`
- Next test in order (#75) fails with `Multiple failures or warnings in test`
- The last test before test execution is halted (#76) fails with `The tested app UniformDocs unexpectedly stopped during the test.`

#### Screenshot

Screenshot 2019-05-02 at 12 49 32

PR #400, opened 2019-04-15
Example implementation for nested view-model with uni-data-table

PatrykSzwer:


src/UniformDocs/Helpers/PersonFilter.cs
```diff
@@ -13,7 +14,10 @@ protected override IQueryable<Person> ApplyFilter(IQueryable<Person> data, Filte
             if (filter == null) throw new ArgumentNullException(nameof(filter));
             if (filter.PropertyName == nameof(Email.Address))
             {
-                return data.Where(person => person.Email.Address.Contains(filter.Value));
+                // Entire string comparison with a method to ignore diacritics.
+                return data.ToList().Where(person => string.Compare(person.Email.Address, filter.Value,
+                                                                      CultureInfo.CurrentCulture, CompareOptions.IgnoreNonSpace) ==
+                                                                  0).AsQueryable();
```
warpech: It makes more sense to ignore diacritics when filtering the first and last name. I think that email addresses should be actually matched exactly. Can you invert this rule perhaps?

PatrykSzwer: > I think that email addresses should be actually matched exactly

But it will be with this setting. Email address in this case **will be** matched exactly.

warpech: I have just checked, and my result is:

- `First name`. Expected: it should be matched with ignored diacritics. Actual: It is matched exactly
- `Email`. Expected: it should be matched exactly. Actual: It is not matched at all, writing anything to the filter input removes all rows

Could you pls double check?

PatrykSzwer: I think I misunderstand what you mean by **exactly**.
For me **exactly** means matching entire string:
repro

PatrykSzwer: But I agree that ignoring diacritics makes more sense in the first name than in the email. (Although its possible to have email with diacritics, now when I think about it?)

warpech: All filters should have partial text matching. 

By exactly, I meant to treat email: "michal@starcounter.com" and "michał@starcounter.com" as two different emails, because the mail server treats it differently. Typing "michał" in the email search field should not find "michal" for the sake of this demo.

PatrykSzwer: I had a little fun with that and proposed a completly new approach that allows to do partial string search without respecting diacritics. Please let me know what do you think and if you agree I'll replace current example in documentation with this one.

src/UniformDocs/Helpers/PersonFilter.cs
```diff
@@ -12,12 +12,18 @@ public class PersonFilter : QueryableFilter<Person>
         protected override IQueryable<Person> ApplyFilter(IQueryable<Person> data, Filter filter)
         {
             if (filter == null) throw new ArgumentNullException(nameof(filter));
+            if (filter.PropertyName == nameof(Person.FirstName))
+            {
+                // Match normalized string without diacritics.
+                return data.ToList().Where(person =>
+                        new string(person.FirstName.Normalize(NormalizationForm.FormD).Where(c => c < 128).ToArray()).Contains(filter.Value) ||
+                        person.FirstName.Contains(filter.Value)) // if someone is using diacritic characters in search
+                    .AsQueryable();
```
warpech: I spent quite a time trying to understand what's happening here:

- Normalizing to form D splits charactes like è to an e and a nonspacing `
- Filtering characters with the ASCII code lower than 128 removes the "nonspacing" characters

I propose to add this or similar explanation as a comment.

src/UniformDocs/Helpers/PersonFilter.cs
```diff
@@ -0,0 +1,45 @@
...
+using System;
+using System.Linq;
+using System.Text;
+using UniformDocs.Database;
+
+namespace UniformDocs.Helpers
+{
+    public class PersonFilter : QueryableFilter<Person>
+    {
+        protected override IQueryable<Person> ApplyFilter(IQueryable<Person> data, Filter filter)
```
warpech: It is hard to test the validity of the filtering, because the `DummyData.cs` does not include diacritics. I propose to change the dummy data to:

```
                        new Person
                        {
                            FirstName = $"{i} mémbęr first name",
                            LastName = $"{i} mémbęr last name",
                            Email = new Email
                            {
                                Address = $"{i}mémbęr@email.com",
                                Type = "Work"
                            }, 
                        };
```

warpech:
@miyconst could you please review the solution to https://github.com/Starcounter/Starcounter.Uniform/issues/19#issuecomment-483691953 presented here?

src/UniformDocs/Helpers/PersonFilter.cs
warpech: I like it but it needs @miyconst's opinion above mine

src/UniformDocs/Helpers/PersonFilter.cs
PatrykSzwer: To also add some diversity and make understanding of the implemented filters easier, I'll add both, string with and without diacritics.

Issue #396, opened 2019-03-04
Implement fully parallelizable testing


warpech:
Right now, the tests in UniformDocs (and Blending, etc) are not parallelizable. I think that in case of UniformDocs, this is because of one problem:

1. Some tests work on data that is commited to the db

I was thinking that if we could change the code so that the tested pages would use long-running transactions and never commit then we could have fully parallelizable tests.

@kegor could you please give it a try? Please don't spend on it more than a day.

warpech:
In BrowserStack, the queue size is 5, so you can't just push all tests at once to the queue. However, you can retry to put the test in a queue if it was rejected.

There is a REST API method that can return the size of the queue: https://www.browserstack.com/automate/rest-api (edited)

kegor:
I would like to discuss few point which i found during the debugging the possible solutions:
1. On the https://automate.browserstack.com i see that the minimum time of test execution is ~15 sec (Chrome 69.0, Win 10: UniformDocs.Tests.Test.CardPageTest). The longest time that i saw is ~2m 30.
2. Following request: `curl -u "marcinwarpechows2:xAaCqzP1xxjpNsN6b6AM" https://api.browserstack.com/automate/plan.json` returns JSON object:
```
{"automate_plan":"Automate Pro","parallel_sessions_running":0,"team_parallel_sessions_max_allowed":2,"parallel_sessions_max_allowed":2,"queued_sessions":0,"queued_sessions_max_allowed":5}
```
I wrote simple method to be possible analyze these values during the tests running. Probably significant note here from the `BrowserStack` docs:
>**Note:** You can make up to 800 API requests per 5 minutes per user.
3. During attempts to use `Threading.Timer` class for waiting while `parallel_sessions_running` becomes `<2` (or `queued_sessions <4`) i found probably obvious thing that each test class running in the separate thread, and, in my understanding, if we would like to wait while one of 2 available parallel threads on `BrowserStack` (according to the current plan) will be free then we should block all these threads and wait significant amount of time while all tests in one class will be executed.

@warpech, @PatrykSzwer maybe you have more clear picture in your mind how this should work and you can share here?

warpech:
I discussed on Slack with @kegor that I think point 2 can be omitted if he just tries to create a RemoteWebDriver session and then, in case of `System.InvalidOperationException : All parallel tests are currently in use, including the queued tests. Please wait to finish or upgrade your plan to add more sessions.`, puts an attempt to retry into a queue on our side.

[...]

PR #402, opened 2019-04-22
[Discussion] Problem 396: fully parallelizable testing.

kegor:
**Problem #396:**
Right now, the tests in UniformDocs (and Blending, etc) are not parallelizable.

**Solution:**
In the first iteration i would like to discuss with you the approach with re-running the tests that leads to throwing the `WebDriverException` with the main message `All parallel tests are currently in use` on the first attempt to create `WebDriver` for them in `TestFixtureSetUp` method.
For now was added very simple implementation that uses .Net queue for collecting such tests and `NUnit.Engine` methods for running them later programmatically.

**Screenshots:**
_Highlighted tests were started from the queue mentioned above:_
image

_Four inconclusive tests are those which where started programmatically later:_
image

**Questions to discuss:**
1. The main question/problem here on my view is marking the programmatically runned tests `Passed` after they `Inconclusive` state, i didn't yet find the straightforward solution yet.
2. Using `Thread.Sleep` with constant parameter is always fragile, during the manual tests i see that it doesn't work with value `10s` even on currently only 5 `[Parallelizable(ParallelScope.Fixtures)]` tests active.
In this case i saw tests (on BrowserStack) with status:
```
 Failed
(Marked via REST API : Inconclusive in setup)
```


test/UniformDocs.Tests/Test/BaseTest.cs
```diff
@@ -49,7 +48,24 @@ public void TestFixtureSetUp()
 
             if (_browsersToRun.Contains(Config.BrowserDictionary[_browser]))
             {
-                Driver = WebDriverManager.StartDriver(_browser, Config.Timeout, serverUri);
+                try
+                {
+                    Driver = WebDriverManager.StartDriver(_browser, Config.Timeout, serverUri);
```
warpech: I have a very stupid idea, but I must ask... 

Could we just do a naive retry using only a `while` and `Thread.Sleep`? Maybe such solution can't work, but if it does, I guess it would solve the need for doing `Assert.Inconclusive`.

```cs
var tries = 0;
while(tries < 5)
{
    try
    {
        Driver = WebDriverManager.StartDriver(_browser, Config.Timeout, serverUri);
    }
    catch (WebDriverException ex)
    {
        if (ex.Message.Contains("All parallel tests are currently in use"))
        {
            Thread.Sleep(30000);
            tries++;
        }
    }
}
if (Driver == null)
{
    throw new Exception("Can't create WebDriver, probably due to a bad config or because all parallel tests are currently in use.");
}
```

PatrykSzwer: Yes, I agree, thats basically what I suggested in the second part of my comment.

kegor: Probably i missed newest changes in Blending tests, let me please take a look.

kegor: Alright guys, i tried this way, i'll paste results here right now for the possibility to think on them in parallel (maybe you'll get the answer faster than me):
image

image

I tried several runs with 10, 20 and 30 seconds as the parameter for the `Thread.Sleep()`, and actually the errors were mostly the same for 20 and 30 (pictures above) and for 10 i had one:
image

that i guess can be answered quickly, with this interval 5 while loops is to little for getting second thread free on BrowserStack.

warpech: I don't know if I understand the problem, are you suggesting that `Thread.Sleep(20 * 1000)` in one test, halts the execution of another test that is already started, causing BrowserStack to time out?

kegor: @warpech, yeah! Now it works without any fails and really faster then before (i tried 2 times), now we can try the same interval (20 sec) with all other uncommented tests right?

test/UniformDocs.Tests/Test/AutoCompletePageTest.cs
```diff
@@ -5,9 +5,10 @@
 namespace UniformDocs.Tests.Test
 {
     [TestFixture(Config.Browser.Chrome)]
-    [TestFixture(Config.Browser.ChromeNoV0)]
+    //[TestFixture(Config.Browser.ChromeNoV0)]
```
warpech: We need `ChromeNoV0` tests to be enabled, but I had a problem to run them using Selenium 3.141.0 in the past. Selenium 3.14.0 works fine.

test/UniformDocs.Tests/Test/BaseTest.cs
warpech: Yes, please try adding `Parallelizable` to all test classes. Also, `ChromeNoV0` needs to be re-enabled.

warpech:
The build failed, do you think this is because the number of retries is too small? https://teamcity.starcounter.org/viewLog.html?buildId=95776&buildTypeId=Bifrost_Webplatform_UniformDocs_IntegrationUniformDocsDebug

I have started another build: https://teamcity.starcounter.org/viewLog.html?buildId=95839&

kegor:
@warpech, alright, how i see the situation to sync my understanding with your:
image
1. We increasing the numbers on this picture calling `Driver = WebDriverManager.StartDriver(_browser, Config.Timeout, serverUri);`
2. When the numbers become `2/2 5/5` and we trying to start next `WebDriver` we getting `WebDriverException` and we forcing to sleep current thread on 1 min.
3. If during this 1 min the numbers become `2/2 4/5` and we will start new `WebDriver` then we could see again `2/2 5/5`, if not we will wait next 1 min (probably `public **async** Task TestFixtureSetUp()` reducing this time because of trying to start driver in multiple threads but this fact isn't affect the main idea).

Summary: considering these steps i could say that we anyway forced to wait until `Queued` queue will free the slot, and we can't then somehow influence on the process of moving these tests from `Queued` to the `Running` are you agree?

warpech:
Yes, that's what I think. I think that the delay should be changed from 60 seconds to a random number in the range 20-60 seconds, otherwise all the retries are going in one huge wave.

I have committed code that adds some diagnostic messages that make it easier to investigate what is happening with the queued tests. Please feel free to add more.

test/UniformDocs.Tests/Test/BaseTest.cs
```diff
@@ -49,7 +49,38 @@ public void TestFixtureSetUp()
...
+                            TestContext.Progress.WriteLine($"{TestContext.CurrentContext.Test.FullName} connection FAIL...");
+                            await Task.Delay(TimeSpan.FromSeconds(60));
+                            TestContext.Progress.WriteLine($"{TestContext.CurrentContext.Test.FullName} connection will retry now");
+                            tries++;
+                        }
+                    }
+
+                    if (Driver != null)
+                    {
+                        break;
```
warpech: This seems to be made redundant by the break on line 61.

warpech:
The build has [passed](https://teamcity.starcounter.org/viewLog.html?buildId=95908&buildTypeId=Bifrost_Webplatform_UniformDocs_IntegrationUniformDocsDebug), which is great! But the execution time is not different than normally. Let's re-run the build at a later hour, when BrowserStack is idle, to see if this was caused by BrowserStack congestion.

kegor:
@warpech, second run took the same 22 minutes.

warpech:
We have a problem with handling the full queue. On Saturday at 20:20 I have started 4 simultaneous builds on different build agents.

All builds (build number 61, 62, 63, 64) failed: https://teamcity.starcounter.org/viewType.html?buildTypeId=Bifrost_Webplatform_UniformDocs_IntegrationUniformDocsDebug&tab=buildTypeHistoryList&branch_Bifrost_Webplatform_UniformDocs=__all_branches__ with many same errors:

```
System.Exception : Can't create WebDriver, probably due to a bad config or because all parallel tests are currently in use.
   at UniformDocs.Tests.Test.BaseTest.<TestFixtureSetUp>d__7.MoveNext() in C:\tc_work\Bifrost_Webplatform_UniformDocs_IntegrationUniformDocsDebug\UniformDocs\test\UniformDocs.Tests\Test\BaseTest.cs:line 80
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NUnit.Framework.Internal.AwaitAdapter.TaskAwaitAdapter.BlockUntilCompleted() in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\AwaitAdapter.cs:line 109
   at NUnit.Framework.Internal.AsyncToSyncAdapter.Await(Func`1 invoke) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\AsyncToSyncAdapter.cs:line 79
   at NUnit.Framework.Internal.Commands.SetUpTearDownItem.RunSetUp(TestExecutionContext context) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Commands\SetUpTearDownItem.cs:line 68
   at NUnit.Framework.Internal.Commands.BeforeTestCommand.Execute(TestExecutionContext context) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Commands\BeforeTestCommand.cs:line 50
   at NUnit.Framework.Internal.Commands.BeforeTestCommand.Execute(TestExecutionContext context) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Commands\BeforeTestCommand.cs:line 50
   at NUnit.Framework.Internal.Commands.BeforeTestCommand.Execute(TestExecutionContext context) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Commands\BeforeTestCommand.cs:line 50
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformOneTimeSetUp() in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Execution\CompositeWorkItem.cs:line 257
```

